PR #9, opened 2022-10-23
F-16I mod v2.2 support (by IDF Mods Project)

MetalStormGhost:
Added a pydcs extension for the IDF Mods Project F-16I mod version 2.2, utilizing the pylon_injector written for the CJS Superbug mod support. Includes banner by Schmokedpancake.

The mod alters the stock F-16C aircraft, which requires a different approach when injecting the mod information to pydcs. The extension changes the F-16C parameters by setattr(F_16C_50, "fuel_max", 2585.48), adds new weapons to the existing pylons and adds five new ones (for example, the mod allows internal ECM equipment in the dorsal spine, freeing the central hardpoint for a fuel tank).

The mod also adds additional Vipers: F-16D (both with and without the dorsal spine, GE and P&W engines) and the stock F-16C Block 50 under a different DCS unit ID.

Also added a separate mod_payloads directory under resources which can be used when aircraft added by mods use the same "unitType" in the loadout lua files. The loadouts are only loaded if the mod is enabled (uses the existing inject_custom_payloads function).

The mod settings are now saved in the Liberation save inside the Faction, so loading the aircraft icons, banners, loadouts etc and the pydcs injection can be re-applied on game load.


Raffson:
Just to track the issues we need to (re)tackle:

- Re-export the F-16I mod, since there's still references to the old livery implementation
- ~~Add the F-16D/I as Anti-Ship capable? Pydcs seems to suggest F-16 in general are capable, but Liberation never added them in `ai_flight_planner_db`, so I'm not sure how to proceed here... @MetalStormGhost any suggestions?~~

PR #24, opened 2022-11-26
Update uk_1990.json

ColonelAkirNakesh:
Replaced USS Tarawa with HMS Invincible. 

Successfully tested replacing Tarawa w Invincible as valid control point on Pacific Repartee with four Harriers parked.

### DIFF
--- a/resources/factions/uk_1990.json
+++ b/resources/factions/uk_1990.json
@@ -55,7 +55,7 @@
     "FFG Oliver Hazard Perry",
     "DDG Arleigh Burke IIa",
     "CG Ticonderoga",
-    "LHA-1 Tarawa",
+    "R05 Invincible"
     "CVN-74 John C. Stennis"
   ],
   "missiles": [],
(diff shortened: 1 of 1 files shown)


resources/factions/uk_1990.json
```diff
@@ -55,7 +55,7 @@
     "FFG Oliver Hazard Perry",
     "DDG Arleigh Burke IIa",
     "CG Ticonderoga",
-    "LHA-1 Tarawa",
+    "R05 Invincible"
```
MetalStormGhost: Comma missing from here?

ColonelAkirNakesh: Thanks, yeah! I removed the Stennis when I tested it but then re-added it but forgot the comma... doh!

MetalStormGhost: Please add the comma, otherwise we can't merge this since it wouldn't be valid json.

ColonelAkirNakesh: D'oh, thanks, fixed. Sorry for being such a nub.

PR #27, opened 2022-12-03
Revert "Revert "Add support for the AI-only F-14A.""

MetalStormGhost:
This reverts commit 920c9b15ff815b5df08ff90c4a6b0a10cce3e169.

Restores the AI F-14A, which allows campaigns to use different air-to-air loadouts for different sides which are both flying the F-14A. Also restores the newer icons for player-flyable F-14 variants, getting rid of the ridiculous black livery which was never used operationally.


MetalStormGhost:
Reverting a revert. I'm awaiting the first "yo dawg" joke.

Raffson:
> Reverting a revert. I'm awaiting the first "yo dawg" joke.

Sounds legit 🤣

PR #116, opened 2023-05-04
Dev sw3

Pande4360:
A rework of the old modpack 2.0 now in 2.55.
Everything has been reworked for full functionality with AI and balancing in context with this dcs-retribution campaign software.
Mod link: https://www.nexusmods.com/dcsworld/mods/7

Adds Aircraft: X-Wing, A-Wing, Y-Wing, N1 Starfighter, Millenium Falcon, Tie Fighter, Tie bomber, Tie Hunter, Tie Interceptor, Raider Corvette
Adds Vehicles: ATST ATAT Juggernaut MBT9s and MBT9 AA Variant
Adds star destroyer, though not yet implemented.
Adds 4 Factions, Rebels and Imperials, which are divided into AI and Player varient for now. That means the Player has to use the player faction and select the AI faction for the CPU. The players can then assign AI variants of aircraft for the AI and select specifically for player designed aircrafts for themselves. The only downside of this is that you cannot have AI wingmates in your flight. Dont worry everything works fine. Ill look into it if I can change this in the future.
Added payloads

All factions are fully capable of SEAD DEAD CAS CAP missions. Not sure about RUNWAY Strikes but who needs that anyway. In any case you can still add native aircrafts to it.


game/ato/ai_flight_planner_db.py
Raffson, line 1: Correct me if I'm wrong, but it looks like you only added the AI variants?
If that's the case then include all "aircraft", otherwise the player variants will never be used by the autoplanner.

Raffson (dismissed):
Aside from ai_flight_planner_db  everything seems to look good 👍

game/ato/ai_flight_planner_db.py
Pande4360: Hey this is done on purpose. As the player variants are only to be used by the player. And they can set up that themselves quickly. I guess I could add them too but that would need the player to configure them as player only, which seems more of a hassle to me.

Raffson: Right, makes sense 👍

Raffson (approved):

PR #125, opened 2023-05-29
Use SEAD task for stock Tornado SEAD/DEAD

MetalStormGhost:
The stock Tornado is capable of the SEAD task in DCS, but not the CAS task, so use SEAD for Tornadoes when flying SEAD or DEAD missions. Also, feel free to add to the list if other airframes should use SEAD as well for cosmetic reasons (would show SEAD task in the mission editor and F10 map in-game).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,7 @@
 * **[UI]** Added fuel selector in flight's edit window.
 * **[Plugins]** Expose Splash Damage's "game_messages" option and set its default to false.
 * **[Mission Generation]** Improved AI SEAD capabilities, allowing for mixed loadouts using Decoys, ARMs & ASMs.
+* **[Mission Generation]** Tornadoes now use SEAD tasking for SEAD/DEAD missions.
 * **[Modding]** Support for A-7E Corsair II (presumed latest available version)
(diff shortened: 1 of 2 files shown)


Raffson (approved):

Adecarcer:
Hi.

F-105D and G should also be SEAD and DEAD capable. In fact the G variant was created with that purpose.

Can they also be included in the list?

Raffson:
> Hi.
> 
> F-105D and G should also be SEAD and DEAD capable. In fact the G variant was created with that purpose.
> 
> Can they also be included in the list?

That would actually break the F-104 because it doesn't natively support the SEAD task. The Tornado was the opposite, it supports SEAD but not CAS, and since we model so regular SEAD & DEAD to use the CAS task, that obviously causes trouble 😂

Adecarcer:
Aren't we confusing F104 Starfighter with F105 Thunderchief?

Enviado desde Outlook para Android<https://aka.ms/AAb9ysg>
________________________________

Raffson:
Omg, you're probably right 😂

EDIT: just verified, same story for F-105

PR #132, opened 2023-06-03
Roadbase and ground spawn support

MetalStormGhost:
Implemented support for roadbases and ground spawn slots at airfields and FOBs. The ground spawn slots can be inserted in campaigns by placing either an A-10A or an AJS37 at a runway or ramp. This will cause an invisible FARP, an ammo dump and a fuel dump to be placed (behind the slot in case of A-10A, to the side in case of AJS37) in the generated campaigns. The ground spawn slot can be used by human controlled aircraft in generated missions. Also allowed the use of the four-slot FARP and the single helipad in campaigns, in addition to the invisible FARP. The first waypoint of the placed aircraft will be the center of a Remove Statics trigger zone (which might or might not work in multiplayer due to a DCS limitation).

image
image

Also implemented three new options in settings:
 - AI fixed-wing aircraft can use roadbases / bases with only ground spawns
   - This setting will allow the AI to use the roadbases for flights and transfers. AI flights will air-start from these bases, since the AI in DCS is not currently able to take off from ground spawns.
 - Spawn trucks at ground spawns in airbases instead of FARP statics
 - Spawn trucks at ground spawns in roadbases instead of FARP statics
   - These settings will replace the FARP statics with refueler and ammo trucks at roadbases. Enabling them might have a negative performance impact.


game/missiongenerator/aircraft/aircraftgenerator.py
```diff
@@ -209,13 +226,6 @@ def create_and_configure_flight(
             ).configure()
         )
 
-        wpt = group.waypoint("LANDING")
-        if flight.is_helo and isinstance(flight.arrival, Fob) and wpt:
-            hpad = self.helipads[flight.arrival].units.pop(0)
-            wpt.helipad_id = hpad.id
-            wpt.link_unit = hpad.id
-            self.helipads[flight.arrival].units.append(hpad)
-
```
Raffson: Where is this being handled now?

game/campaignloader/mizcampaignloader.py
```diff
@@ -425,7 +442,66 @@ def add_preset_locations(self) -> None:
...
+                    PointWithHeading.from_point(
+                        static.position, Heading.from_degrees(static.units[0].heading)
+                    )
+                )
+            else:
+                closest.helipads_invisible.append(
+                    PointWithHeading.from_point(
+                        static.position, Heading.from_degrees(static.units[0].heading)
+                    )
+                )
```
Raffson: Refactor to a (private) method `add_helipad`?

game/campaignloader/mizcampaignloader.py
```diff
@@ -425,7 +442,66 @@ def add_preset_locations(self) -> None:
...
+
+            closest.ground_spawns_roadbase.append(
+                (
+                    PointWithHeading.from_point(
+                        plane_group.position,
+                        Heading.from_degrees(plane_group.units[0].heading),
+                    ),
+                    first_waypoint,
+                )
+            )
```
Raffson: Refactor to `add_roadbase` ?

game/campaignloader/mizcampaignloader.py
```diff
@@ -425,7 +442,66 @@ def add_preset_locations(self) -> None:
...
+
+            closest.ground_spawns.append(
+                (
+                    PointWithHeading.from_point(
+                        plane_group.position,
+                        Heading.from_degrees(plane_group.units[0].heading),
+                    ),
+                    first_waypoint,
+                )
+            )
```
Raffson: Refactor to `add_roadbase_parking_slot` ?

game/missiongenerator/aircraft/aircraftgenerator.py
```diff
@@ -66,6 +69,8 @@ def __init__(
         self.flights: List[FlightData] = []
         self.mission_data = mission_data
         self.helipads = helipads
+        self.ground_spawns_roadbase = ground_spawns_roadbase
+        self.ground_spawns = ground_spawns
```
Raffson, line 74: Would it make sense to refactor this into a container class? Something like `CustomAirport` ?

game/missiongenerator/aircraft/flightgroupspawner.py
```diff
@@ -47,13 +47,17 @@ def __init__(
...
+        ground_spawns_roadbase: dict[ControlPoint, list[Tuple[StaticGroup, Point]]],
+        ground_spawns: dict[ControlPoint, list[Tuple[StaticGroup, Point]]],
         mission_data: MissionData,
     ) -> None:
         self.flight = flight
         self.country = country
         self.mission = mission
         self.helipads = helipads
+        self.ground_spawns_roadbase = ground_spawns_roadbase
+        self.ground_spawns = ground_spawns
```
Raffson, line 60: Looks like a refactor would definitely be a good idea here 😆

game/missiongenerator/aircraft/flightgroupspawner.py
```diff
@@ -283,6 +375,12 @@ def dcs_start_type(self) -> DcsStartType:
             return DcsStartType.Warm
         raise ValueError(f"There is no pydcs StartType matching {self.start_type}")
 
+    def _start_type_at_airfield(
+        self,
+        airfield: Airfield,
```
Raffson, line 381: Unused?

game/missiongenerator/missiongenerator.py
```diff
@@ -226,6 +226,8 @@ def generate_air_units(self, tgo_generator: TgoGenerator) -> None:
             self.unit_map,
             mission_data=self.mission_data,
             helipads=tgo_generator.helipads,
+            ground_spawns_roadbase=tgo_generator.ground_spawns_roadbase,
+            ground_spawns=tgo_generator.ground_spawns,
```
Raffson, line 247: More lines that would be saved by the refactor 😂

game/missiongenerator/tgogenerator.py
```diff
@@ -9,13 +9,107 @@
...
-from typing import Any, Dict, Iterator, List, Optional, TYPE_CHECKING, Type
+from collections import defaultdict
+from typing import Any, Dict, Iterator, List, Optional, TYPE_CHECKING, Type, Tuple
 
 import dcs.vehicles
-from dcs import Mission, Point
+from dcs import Mission, Point, unitgroup
 from dcs.action import DoScript, SceneryDestructionZone
 from dcs.condition import MapObjectIsDead
+from dcs.countries import (
```
Raffson: `from dcs.countries import *`perhaps?

game/game.py
```diff
@@ -90,6 +90,8 @@ class TurnState(Enum):
 
 
 class Game:
+    scenery_clear_zones: List[Point]
```
Raffson, line 93: I guess this can also be included in the changelog, since campaign designers will have the ability in general... though keep in mind to mention that it doesn't work properly in MP

game/procurement.py
```diff
@@ -232,11 +237,15 @@ def best_squadrons_for(
...
+            parking_type = ParkingType().from_squadron(squadron)
+
             if not squadron.can_provide_pilots(request.number):
                 continue
+            if squadron.location.unclaimed_parking(parking_type) < request.number:
+                continue
             if not squadron.has_aircraft_capacity_for(request.number):
                 continue
-            if squadron.location.unclaimed_parking() < request.number:
+            if squadron.location.unclaimed_parking(parking_type) < request.number:
```
Raffson: I this second check intentional?

game/theater/controlpoint.py
```diff
@@ -757,17 +812,22 @@ def aircraft_retreat_destination(
         self, squadron: Squadron
     ) -> Optional[ControlPoint]:
         closest = ObjectiveDistanceCache.get_closest_airfields(self)
-        max_retreat_distance = squadron.aircraft.max_mission_range
+        # Multiply the max mission range by two when evaluating retreats,
+        # since you only need to fly one way in that case
+        max_retreat_distance = squadron.aircraft.max_mission_range * 2
```
Raffson, line 821: Add "Improved squadron retreat logic" to changelog?

qt_ui/windows/basemenu/airfield/QAircraftRecruitmentMenu.py
```diff
@@ -91,8 +91,13 @@ def __init__(self, game_model: GameModel, control_point: ControlPoint) -> None:
         self.setAlignment(Qt.AlignLeft)
 
     def update_label(self) -> None:
-        next_turn = self.control_point.allocated_aircraft()
-        max_amount = self.control_point.total_aircraft_parking
+        parking_type = ParkingType()
+        parking_type.include_rotary_wing = True
+        parking_type.include_fixed_wing = True
+        parking_type.include_fixed_wing_stol = True
```
Raffson: I see this part returning a lot... perhaps there's a cleaner way to initialize this with the constructor?

Raffson (changes requested):

game/missiongenerator/aircraft/aircraftgenerator.py
MetalStormGhost: Maybe, but I don't want to do it here at this stage. The ground spawns are handled identically to helipads right now, but I'm not sure if that container class should only contain helipads and ground spawns?

game/procurement.py
MetalStormGhost: Yes. has_aircraft_capacity_for() doesn't take into account the parking type, it only cares about the number. unclaimed_parking(), on the other hand, takes the parking type into account.

However, should I refactor has_aircraft_capacity_for() to actually care about the parking type?

game/missiongenerator/aircraft/aircraftgenerator.py
Raffson: That would make sense but yeah, if you want to keep this for later I'm all good with that 👍

game/procurement.py
Raffson: Yeah it looks quite similar but I wouldn't refactor unless has_aircraft_capacity_for is only used here, but I doubt that's the case 😅

game/campaignloader/mizcampaignloader.py
MetalStormGhost: Done.

game/campaignloader/mizcampaignloader.py
MetalStormGhost: Done.

game/campaignloader/mizcampaignloader.py
MetalStormGhost: Done.

game/game.py
MetalStormGhost: Yeah. Good point. This has now been added to the changelog.

game/theater/controlpoint.py
MetalStormGhost: Added.

qt_ui/windows/basemenu/airfield/QAircraftRecruitmentMenu.py
MetalStormGhost: There actually already was, I was just not using the contructor for that. Personal preference, I guess. I have now changed it as you suggest, and it does look cleaner.

game/missiongenerator/tgogenerator.py
MetalStormGhost: Done.

game/missiongenerator/aircraft/aircraftgenerator.py
MetalStormGhost: Same place. It's now been restored.

game/missiongenerator/aircraft/flightgroupspawner.py
MetalStormGhost: It's used here:

>     def _generate_at_airfield(self, name: str, airfield: Airfield) -> FlyingGroup[Any]:
>         # TODO: Delayed runway starts should be converted to air starts for multiplayer.
>         # Runway starts do not work with late activated aircraft in multiplayer. Instead
>         # of spawning on the runway the aircraft will spawn on the taxiway, potentially
>         # somewhere that they don't fit anyway. We should either upgrade these to air
>         # starts or (less likely) downgrade to warm starts to avoid the issue when the
>         # player is generating the mission for multiplayer (which would need a new
>         # option).
>         self.flight.unit_type.dcs_unit_type.load_payloads()
>         return self.mission.flight_group_from_airport(
>             country=self.country,
>             name=name,
>             aircraft_type=self.flight.unit_type.dcs_unit_type,
>             airport=airfield.airport,
>             maintask=None,
>             start_type=self._start_type_at_airfield(airfield),
>             group_size=self.flight.count,
>             parking_slots=None,
>         )

Raffson (approved):

PR #155, opened 2023-07-01
WRL Faction Updates

holyorangejuice:
I've updated all WRL factions and added a new WRL faction. All factions now share the exact same air lineup so they can be used anywhere. When I design a mission, I still have to use custom factions to make sure the campaign is played as planned. A player, using the UI, can customize that experience. As a mission designer I can't use one faction and define which units are on or off in the UI. So until then, multiple factions. 

The new WRL mixed faction is going to be used in all wargame scenarios and some Sinai map stuff I've been looking at making. Also the file names are annoying me because they are not how I typed them. Some have CAPS, some have lowercase...

WRL_TaskForceBlue
WRL_TaskForceInsurgent
WRL_TaskForceMixed
WRL_TaskForceRed


Raffson (approved):

Raffson:
Windows is case-insensitive when it comes to paths, so the trick is to rename it to something else first, and then rename it back to what you would want it to be.

PR #160, opened 2023-07-03
SEAD Sweep threat buffer distance

MetalStormGhost:
Added a new option in settings: SEAD Sweep threat buffer distance (nmi).


Raffson:
Wouldn't it make more sense to use this setting for the SEAD Sweep's engagement distance? That default threat range of 5NM is simply to cope with packages where the target doesn't have a threat range, in which case it would make sense for SEAD (Sweep) flights to fly in closer.

Raffson (approved):
That works too I guess 😅

PR #177, opened 2023-07-19
F-104 missing weapons

Turbolious:
Added the 2x Aim-9p/L variants and the Aspide MRAAM


Raffson (approved):
While we currently don't support the Aspide at all, it's good to have it already in place for when we make the necessary adjustments to support the missile.

PR #227, opened 2023-12-28
AI unlimited fuel initial implementation

tmz42:
AI unlimited fuel implementation for DCS 2.9.

For player flights, adds command at first waypoint.

For AI-only flights, adds command at join and removes it at split.

### DIFF
--- a/game/missiongenerator/aircraft/flightgroupconfigurator.py
+++ b/game/missiongenerator/aircraft/flightgroupconfigurator.py
@@ -7,7 +7,7 @@ from typing import Any, Optional, TYPE_CHECKING
 from dcs import Mission, Point
 from dcs.action import DoScript
 from dcs.flyingunit import FlyingUnit
-from dcs.task import OptReactOnThreat, SetUnlimitedFuelCommand
+from dcs.task import OptReactOnThreat
 from dcs.translation import String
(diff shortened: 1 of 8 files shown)


game/missiongenerator/aircraft/flightgroupconfigurator.py
```diff
@@ -126,6 +126,13 @@ def configure(self) -> FlightData:
             self.flight.flight_plan.waypoints,
         )
 
+        # Unlimited fuel option : enable on player flights. Must be first option to work.
+        if (
+            self.flight.squadron.coalition.game.settings.ai_unlimited_fuel
+            and self.flight.client_count
+        ):
+            self.group.points[0].tasks.insert(0, SetUnlimitedFuelCommand(True))
+
```
Raffson: This part should probably move down, removing the need to use 'insert'

tmz42:
The section moved to Aircraft Behavior, I did not notice your comment about insert before commiting, I generalized it as it is preferred for the UnlimitedFuel command to be the first one. Do you want me to use add instead?

Otherwise it should be OK, I checked the behaviour and it works both for player flights, and allied and OPFOR flights with join/split.

Raffson (approved):

PR #233, opened 2023-12-30
IDF Assets Pack support v1.1

MetalStormGhost:
Updated Iron Dome support to IDF Assets Pack V1.1, which includes the David's Sling. Added an Israel 2017 faction with the David's Sling.

@Raffson I have now implemented this in a way which replaces the Iron Dome pydcs export. Do we need to support both mods simultaneously? I've understood that the IDF Assets Pack is an evolution of the Iron Dome mod, but it might be possible to make them both being supported at the same time with some added complexity (some of the unit IDs overlap between the mods). I'm just not sure if that added complexity is justified.


Raffson:
Considering the fact that some of the units were deleted, we might need to deal with this in the migrator to keep supporting campaigns that used the Iron Dome mod...

Raffson (approved):

PR #249, opened 2024-02-11
S3B tasks update

Starfire13:
The reworked S-3B now has greater capabilities and is suitable for DEAD, BAI, and CAS tasking. I have also included a default loadout file that uses its new weapons.


Starfire13:
OK, something else is blocking this change.

Screenshot 2024-02-11 162827

Raffson (approved):

PR #276, opened 2024-03-24
Small update for Scenic Route I + Tomcat Livery Sets

tmz42:
Updated Scenic Route for a factory at FOB I and replaced FOB by airport
Added Livery Sets for some Tomcat Squadrons, and added Allied Force Black Aces and Tophatters (thanks for this one @Raffson !)


Raffson:
Gotta thank @Adecarcer for that one since it was his idea 😅

Raffson (approved):

PR #277, opened 2024-03-25
CJS super hornet mod update

Chilli935:
- Updated fa18efg pydcs file to mod version 2.2.5 (changes can be found in the CJS discord [here](https://discord.gg/aSN4s2FmHZ))
- Created australia_2009.json
- Removed EA-18G Growler from usn_2009.json
- Created usn_2009.json
- Updated tasks in EA-18G.yaml

### DIFF
--- a/pydcs_extensions/fa18efg/fa18efg.py
+++ b/pydcs_extensions/fa18efg/fa18efg.py
@@ -247,133 +247,133 @@ class FA_18E(PlaneType):
     livery_name = "FA-18E"  # from type
 
     class Pylon1:
-        AIM_9M_Sidewinder_IR_AAM = (1, WeaponsFA18EFG.AIM_9M_Sidewinder_IR_AAM)
-        CATM_9M = (1, WeaponsFA18EFG.CATM_9M)
+        AIM_9M_Sidewinder_IR_AAM = (1, Weapons.AIM_9M_Sidewinder_IR_AAM)
+        CATM_9M = (1, Weapons.CATM_9M)
         # ERRR {CATM-9M}
         # ERRR {CATM-9X}
-        AIM_9X_Sidewinder_IR_AAM = (1, WeaponsFA18EFG.AIM_9X_Sidewinder_IR_AAM)
+        AIM_9X_Sidewinder_IR_AAM = (1, Weapons.AIM_9X_Sidewinder_IR_AAM)
         AN_ASQ_T50_TCTS_Pod___ACMI_Pod = (
             1,
-            WeaponsFA18EFG.AN_ASQ_T50_TCTS_Pod___ACMI_Pod,
+            Weapons.AN_ASQ_T50_TCTS_Pod___ACMI_Pod,
         )
 
     class Pylon2:
-        TLAU_127 = (2, WeaponsFA18EFG.TLAU_127)
+        TLAU_127 = (2, Weapons.TLAU_127)
         AIM_120D_AMRAAM___Active_Rdr_AAM = (
             2,
             WeaponsFA18EFG.AIM_120D_AMRAAM___Active_Rdr_AAM,
         )
         AIM_120C_AMRAAM___Active_Radar_AAM = (
             2,
-            WeaponsFA18EFG.AIM_120C_AMRAAM___Active_Radar_AAM,
+            Weapons.AIM_120C_AMRAAM___Active_Radar_AAM,
         )
         # ERRR {CATM-120C}
         LAU_115C_with_AIM_7M_Sparrow_Semi_Active_Radar = (
             2,
-            WeaponsFA18EFG.LAU_115C_with_AIM_7M_Sparrow_Semi_Active_Radar,
+            Weapons.LAU_115C_with_AIM_7M_Sparrow_Semi_Active_Radar,
         )
         LAU_115C_with_AIM_7F_Sparrow_Semi_Active_Radar = (
             2,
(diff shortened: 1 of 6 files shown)


pydcs_extensions/fa18efg/fa18efg.py
```diff
@@ -156,11 +255,8 @@ class Pylon1:
         AN_ASQ_T50_TCTS_Pod___ACMI_Pod = (1, Weapons.AN_ASQ_T50_TCTS_Pod___ACMI_Pod)
 
     class Pylon2:
-        TLAU_127 = (2, WeaponsFA18EFG.TLAU_127)
-        AIM_120B_AMRAAM___Active_Radar_AAM = (
-            2,
-            Weapons.AIM_120B_AMRAAM___Active_Radar_AAM,
-        )
+        TLAU_127 = (2, Weapons.TLAU_127)
```
Raffson: This weapon is actually defined in this file, so it _shouldn't_ reference pydcs' `Weapons` class. The idea is to keep pydcs clean of any mods, which I believe is still the case so this could cause trouble 😬

Can you start by reverting changes where `WeaponsF18EFG` was replaced with `Weapons`?

Raffson (changes requested):

pydcs_extensions/fa18efg/fa18efg.py
Chilli935: My apologies, I recently finished support for the mod in liberation and, being aware that the version in retribution is outdated, copied over the files from my PR. Changes should be reverted in last push.

Raffson: You seem to have replaced _all_ occurrences, but it should only be applied to the weapons that are part of the mod, i.e. weapons that you defined at the start of the pydcs extension 😅

I recommend looking at one of the other mods to understand when to use `Weapons<Mod>` and when to use `Weapons`.

Bottom line, any stock DCS weapon should be referred to from the `Weapons` class while modded weapons use the custom defined weapons class in the pydcs extension.

Let me know if I can be of any further assistance 😉

Chilli935: I believe the classes are correctly referred to now

Chilli935: Just got it and properly changed it, thanks

pydcs_extensions/fa18efg/fa18efg.py
```diff
@@ -228,9 +339,10 @@ class Pylon2:
             2,
             Weapons.GBU_54_V_1_B___LJDAM__500lb_Laser__GPS_Guided_Bomb_LD,
         )
-        AGM_154A___JSOW_CEB__CBU_type_ = (2, Weapons.AGM_154A___JSOW_CEB__CBU_type_)
-        AGM_154C___JSOW_Unitary_BROACH = (2, Weapons.AGM_154C___JSOW_Unitary_BROACH)
```
Raffson, line 232: Does the super-hornet no longer support the JSOWs on these pylons? 😅

Raffson:
~~According to the tests there's still some type-errors left~~ 🤔

pydcs_extensions/fa18efg/fa18efg.py
```diff
@@ -153,13 +252,16 @@ class Pylon1:
...
-        AN_ASQ_T50_TCTS_Pod___ACMI_Pod = (1, Weapons.AN_ASQ_T50_TCTS_Pod___ACMI_Pod)
+        AN_ASQ_T50_TCTS_Pod___ACMI_Pod = (
+            1,
+            Weapons.AN_ASQ_T50_TCTS_Pod___ACMI_Pod,
+        )
 
     class Pylon2:
         TLAU_127 = (2, WeaponsFA18EFG.TLAU_127)
-        AIM_120B_AMRAAM___Active_Radar_AAM = (
+        AIM_120D_AMRAAM___Active_Rdr_AAM = (
```
Raffson: Are you sure we need to replace the B with the D model? Perhaps we need to add them both?

The main reason I think adding both would make sense is an attempt at backwards-compatibility with earlier versions of the superbug, because iirc the B model would be used as a D model, but I believe @MetalStormGhost can shed more light on this 😅

pydcs_extensions/fa18efg/fa18efg.py
Chilli935: Neither pylon 2 nor pylon 2 support the JSOWs, this is probably because of the pylon limitations.

pydcs_extensions/fa18efg/fa18efg.py
Raffson: I consulted with Ghosti and we believe this change should also be reverted. The CLSIDs for the B & D models are the same, and so this would "overwrite" the B model altogether.

I believe that's why we left the D model out, but choosing a B model will eventually result in a D model in-game if I understood correctly.

Chilli935: I've reverted the changes in the last push

resources/factions/australia_2009.json
```diff
@@ -0,0 +1,70 @@
...
+    "LHA-1 Tarawa",
+    "L02 Canberra",
+    "HMAS HOBART DDG39"
+  ],
+  "missiles": [],
+  "air_defense_units": [
+    "SAM Hawk SR (AN/MPQ-50)",
+    "HEMTT C-RAM Phalanx"
+  ],
+  "requirements": {
```
Raffson, line 53: We don't actually support this field, but it shouldn't cause any issues afaik so no need to remove this...

Raffson (approved):

PR #291, opened 2024-04-28
Add C-130J squadron to Operation Noisy Cricket for paradrops

Starfire13:


resources/campaigns/operation_noisy_cricket.yaml
```diff
@@ -54,6 +54,12 @@ squadrons:
       size: 4
   #Al Minhad AFB (61)
   12:
+    - primary: Transport
+      secondary: any
+      aircraft:
+        - C-130J-30 Super Hercules
+        - C-130
+      size: 2
```
Raffson, line 62: Wouldn't it be better to split them into two separate squadrons? Right now you have no guarantee the Herc will be used...

Starfire13: What do you mean? If the user has the C-130J, it'll use that. If not, it'll use the regular C-130 as a fallback. In which case the user will just get a regular transport squadron.

Raffson: Right, that's an oversight on my part. Apologies 😅

Raffson (approved):

PR #303, opened 2024-05-03
Default start type for player flights

MetalStormGhost:
Implemented a new option in settings: Default start type for Player flights.

This might help with some campaigns in the Sinai terrain with the sketchy parking spawns and AI pathfinding on the airfields.


qt_ui/windows/mission/flight/QFlightCreator.py
```diff
@@ -241,10 +247,12 @@ def on_squadron_changed(self, index: int) -> None:
         self.roster_editor.replace(None, None)
         if squadron is not None:
             self.roster_editor.replace(
-                squadron, FlightRoster(squadron, self.flight_size_spinner.value())
```
Raffson: I guess this is where Drexyl's exception is coming from...

qt_ui/windows/mission/flight/QFlightCreator.py
```diff
@@ -255,3 +263,22 @@ def update_max_size(self, available: int) -> None:
...
+        roster = self.roster_editor.roster
+
+        if roster is not None and roster.player_count > 0:
+            start_type = self.game.settings.default_start_type_client
+        else:
+            start_type = self.game.settings.default_start_type
+
+        for i, st in enumerate([b for b in ["Cold", "Warm", "Runway", "In Flight"]]):
+            if start_type.value == st:
+                self.start_type.setCurrentIndex(i)
```
Raffson: Perhaps `self.start_type.setCurrentText(start_type.value)` is a better approach? 😅

qt_ui/windows/mission/flight/settings/QFlightSlotEditor.py
```diff
@@ -131,6 +136,8 @@ def on_player_toggled(self, checked: bool) -> None:
         pilot.player = checked
         self.player_toggled.emit()
 
+        self.pilots_changed.emit()
+
     def on_pilot_changed(self, index: int) -> None:
```
Raffson, line 141: `on_pilot_changed` should emit `pilots_changed` in its `finally` block, otherwise the start-type isn't updated if you have a single client pilot which you switch to a non-client pilot...

game/ato/flightroster.py
```diff
@@ -22,6 +22,10 @@ def squadron(self) -> Squadron:
     def iter_pilots(self) -> Iterator[Pilot | None]:
         yield from self.pilots
 
+    @property
+    def player_count(self) -> int:
```
Raffson, line 26: My preference would be to add a `has_clients` method which will use the same statement as you currently have here, expect for using `any` instead of `len`, although I doubt that the performance gain will be worth it so I'll leave it up to you to decide on that...

qt_ui/windows/mission/flight/QFlightCreator.py
```diff
@@ -32,6 +32,7 @@
 
 class QFlightCreator(QDialog):
     created = Signal(Flight)
+    pilots_changed = Signal()
```
Raffson: I believe it would be a better idea to drop this down to ~~`FlightRosterEditor`~~ `QFlightSlotEditor` (I believe that's the class that's furthest "down the rabbit hole" if I haven't missed anything), and then use that signal from `self.roster_editor.pilots_changed` instead, that way we don't create a dependency on a Signal object in `FlightRosterEditor`'s constructor. Does that make sense? Or have I missed something? 😅

qt_ui/windows/mission/flight/settings/QFlightStartType.py
```diff
@@ -43,6 +50,29 @@ def __init__(self, package_model: PackageModel, flight: Flight):
...
+                self.flight.coalition.game.settings.default_start_type_client
+            )
+        else:
+            self.flight.start_type = (
+                self.flight.coalition.game.settings.default_start_type
+            )
+
+        for i, st in enumerate([b for b in ["Cold", "Warm", "Runway", "In Flight"]]):
+            if self.flight.start_type.value == st:
+                self.start_type.setCurrentIndex(i)
```
Raffson: This can probably be replaced as well by `self.start_type.setCurrentText(self.flight.start_type.value)`

qt_ui/windows/mission/flight/settings/QGeneralFlightSettingsTab.py
```diff
@@ -21,6 +21,7 @@
 
 class QGeneralFlightSettingsTab(QFrame):
     flight_size_changed = Signal()
+    pilots_changed = Signal()
```
Raffson: This one could be removed if we would define a signal in `QFlightSlotEditor`, and then use that signal to pass on to `QFlightStartType`, but even that could be circumvented by doing something like:
```
self.flight_slot_editor = QFlightSlotEditor(package_model, flight, game.game)
qfst = QFlightStartType(package_model, flight)
self.flight_slot_editor.pilots_changed.connect(qfst.on_pilot_selected)
```

Raffson (changes requested):

qt_ui/windows/mission/flight/settings/QFlightStartType.py
MetalStormGhost: Done 👍

qt_ui/windows/mission/flight/settings/QGeneralFlightSettingsTab.py
MetalStormGhost: Done 👍

qt_ui/windows/mission/flight/QFlightCreator.py
MetalStormGhost: Done, I think. Let me know if I missed the old pilots_changed signal somewhere?

game/ato/flightroster.py
MetalStormGhost: This method is an original one from Liberation, which has been refactored out at some point, I simply reintroduced it. This leads me to believe that it migh have had some use previously and I would prefer to keep it as is, since it might become useful again further down the line. Also, as you said, the performance gain would not be significant.

qt_ui/windows/mission/flight/settings/QFlightSlotEditor.py
MetalStormGhost: Done 👍

qt_ui/windows/mission/flight/QFlightCreator.py
MetalStormGhost: Done 👍

qt_ui/windows/mission/flight/settings/QGeneralFlightSettingsTab.py
Raffson: Made a small adjustment after your changes, we can also get rid of the Signal parameter in `QFlightStartType` and connect the signal in `QGeneralFlightSettingsTab`

qt_ui/windows/mission/flight/QFlightCreator.py
Raffson: Looks good 👍

Raffson (approved):

PR #309, opened 2024-05-07
Ramon Airbase hack

MetalStormGhost:
Extend option (so it can be disabled when fixed in DCS) to force air-starts (except for the slots that work) at Ramon Airbase, similar to the Nevatim fix in Retribution 1.3.0. Affected slots are: [1, 2, 3, 4, 5, 6, 13, 14, 15, 16, 17, 18, 61]

https://discordapp.com/channels/1015931619187621999/1025837223821705236/1237373245250994267

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@
 * **[Campaign Setup]** Allow adjustments to naval TGOs (except carriers) on turn 0
 * **[Campaign Design]** Ability to configure specific carrier names & types in campaign's yaml file 
 * **[Mission Generation]** Ability to inject custom kneeboards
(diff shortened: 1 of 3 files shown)


Raffson:
The option in the settings should probably be renamed as well, right? 😅

MetalStormGhost:
> The option in the settings should probably be renamed as well, right? 😅

Agreed! This has now been done. I also fixed a logic error I made earlier.

Raffson:
The description of the setting is still referring to the slots at Nevatim 😅

MetalStormGhost:
> The description of the setting is still referring to the slots at Nevatim 😅

Hmm, could you re-check? This should've been pushed already at the time you posted the message.
image

Raffson (approved):

PR #315, opened 2024-05-19
Allied Sword + faction files updates

tmz42:
# Allied Sword
Use inline faction format for Allied Sword and delete json faction files
Update number of aircraft at Ramat David following Syria update

# Minor faction alterations
Added BS3 or A-10CII to some older factions


resources/factions/Israel-USN_2005_Allied_Sword.json
Raffson, line 1: I'm not sure if it's such a good idea to remove the original faction files, despite having moved them to into the campaign's yaml.

Have you verified that those factions are not used in any other campaign?

tmz42: I checked, both factions are just in use by this campaign.

We could keep the files, but merging some single-use faction files into the campaigns would limit a bit faction creep (I count 116 at the moment with those two removed). The other benefit is that it allows for a single campaign file (except for the version number) for both Retribution and Liberation (currently it's in this format in Liberation).

I plan to do this for Retake the Falklands as well if you agree.

Raffson: If it makes maintenance easier on both ends, then I tend to agree 👍

Raffson (approved):

PR #338, opened 2024-06-09
Update usa_1970.json

ColonelAkirNakesh:
Fixes faction error caused by LARC-V naming

Pull requests should be made against the `dev` branch. Any backports
necessary will be handled by the development team.

Pull requests should be focused on one task. Multiple bug fixes should be
multiple PRs, or at the very least split over multiple (isolated) commits.
We cannot merge half a PR, and combined PRs are more
difficult to review. PRs that do not adhere to this may have their review
delayed.

Prefer rebase to merge, and squash commits as needed to preserve a readable
commit history. This project maintains linear history in the `dev` branch, so
we will either rebase or squash your PR when merging. It is much easier for us
if your branch already has a readable commit history (ensure that your commit
subject lines are clear enough to identify the patch in the git log). An
exception to this is made for large PRs that are likely to require multiple
rounds of review; in that case it's easier if you **don't** do this (GitHub
does not preserve the history of old commits, so we cannot filter a PR for only
new changes if a branch is force pushed) and we will squash it when merging.

New features and bug fixes are usually worth mentioning in the changelog.
Exceptions are fixes for bugs that never shipped (were only present in a canary
build), and changes with no intended user observable behavior, such as a
refactor. If you're comfortable writing the note yourself, add it to
`changelog.md` in the root of the project in the section for the upcoming
release.

### DIFF
--- a/resources/factions/usa_1970.json
+++ b/resources/factions/usa_1970.json
@@ -43,7 +43,7 @@
   ],
   "logistics_units": [
     "Truck M818 6x6",
-    "LARC-V"
+    "LARC-V Amphibious Cargo Vehicle"
   ],
   "infantry_units": [
     "Infantry M249",


Starfire13:
No, the current name is the correct one.

Raffson:
> No, the current name is the correct one.

Indeed, refer to https://github.com/dcs-retribution/dcs-retribution/blob/dev/resources/units/ground_units/LARC-V.yaml

@ColonelAkirNakesh which error are you getting exactly?

ColonelAkirNakesh:
Hey gents, I think I found the problem. In the Pretense alpha branch I was using, it's still called LARC-V Amphibious Cargo Vehicle instead of just LARC-V.

Sorry for my noobiness, dudes! For future reference should I be testing/updating from the latest in #Development?

Starfire13:
It has been called LARC-V in both Liberation and Retribution since the unit was included. If @MetalStormGhost's Pretense branch has a different name, then that is perhaps a problem that may need to be fixed on his end. Possibly, he needs to either change the name in the branch, or add in a second variant matching the name used in Retribution.

Edit: It looks like Ghosti created the unit yaml for the LARC-V at a point when Retribution didn't have one yet. I added the one in Retribution a little over a week after Ghosti made his. But he only added it to the pretense generator branch, not to the main Retribution dev branch. That's why the two have different names. I simply added the same file that I had previously added to Liberation, where it is called LARC-V.

Raffson:
> Hey gents, I think I found the problem. In the Pretense alpha branch I was using, it's still called LARC-V Amphibious Cargo Vehicle instead of just LARC-V.
> 
> 
> 
> Sorry for my noobiness, dudes! For future reference should I be testing/updating from the latest in #Development?

Well, that depends on what you're trying to test 😅
If you're testing the Pretense builds, then you would technically want to fork Ghosti's repo and make a PR towards his branch. But in general, the "dev" branch is what we reference, which is also the default branch.

PR #346, opened 2024-06-22
OH-6 Cayuse mod v1.2 support

MetalStormGhost:
Implemented support for OH-6 Cayuse mod version 1.2 by Tobsen & Eightball. Includes support for the Vietnam Asset Pack v1.0 and loadouts by @Starfire13.

OH-6A
image
image
image
image


resources/units/aircraft/OH-6A.yaml
```diff
@@ -0,0 +1,18 @@
...
+manufacturer: Hughes Helicopters
+origin: USA
+price: 3
+role: Light Observation/Light Attack
+variants:
+  OH-6A Cayuse: {}
+tasks:
+  BAI: 200
+  CAS: 200
+  OCA/Aircraft: 200
```
Raffson: Does the Recon task have the ability to use AttackGroup & SeachAndAttack waypoint tasks? Otherwise this might not work...

Raffson: Also, if it actually has these tasks, then we need to add Recon accordingly in `aircraftbehavior.py`

MetalStormGhost: I'm not sure if that is relevant in the case of this mod, since I saw somewhere (a reference which I can't find right now) that the OH-6 might not work on AI hands at all, or might work poorly...

Raffson: Well then we definitely need to add the fallbacks either way because even for player flights I believe it will run into an exception during generation of a CAS/BAI/OCA mission.

Raffson: Just checked in DCS, seems like Reconnaissance doesn't have any waypoint-task to attack, except strafing but that won't help us I'm afraid...

I would replace the tasks with Transport only, although I guess it's somewhat unrealistic because I can't imagine the little bird hauling tanks, but that's about the only auto-generated mission I would allow 😅

At the same time we would _need_ these tasks because otherwise you can't plan those flights manually for players either, but I would opt for something that checks whether it's an OH-6 & it's a player flight, inject the Reconnaissance task for `configure_cas` & `configure_oca_strike`, otherwise don't inject anything and let the exception throw...

Raffson (approved):

Raffson (approved):

PR #349, opened 2024-06-28
Added Carstens Arty Script

Turbolious:
Added ability to use Carstens Arty Script.

To mark a target location for the artillery, you can place an empty marker on the F10 map, based either on range and bearing to the target or judging by landmarks.
The marker position will then be interpreted as the target location.
In more modern aircraft, that provide the target coordinate, you can enter a MGRS coordinate into the marker's textfield, this will then override the marker's position and will be used instead.
Once a target location is marked, you request a calibration round via the F10 radio menu. In case the fire is off the target, you can adjust the distance and direction and either request another calibration round or request fire for effect.

To prevent exploitation of the script, a maximum distance from player to target is implemented.

Variables like the ammount of rounds to be fired, radius in which the rounds will impact, the time it takes for the rounds to arrive and the maximum target distance can be changed by the user to his liking.


resources/plugins/arty/CG_ArtySpotter_V1_2a_MP.lua
```diff
@@ -0,0 +1,715 @@
...
+local version = "MP 1.2"
+
+-- User configurable variables
+
+local user_fireDelay = 10				-- time to impcat of the rounds
+local user_quantity = 20				-- how many rounds will be fired in a fire for effect task
+local user_spread = 50					-- impact radius of the rounds during fire for effect
+local user_spottingDistance = 15		-- max allowable distance from player to target to prevent cheating. In kilometers.
+local user_restrictByType = ""        	-- Restriction by type ("", "helo", etc.)
+local user_restrictByUnitName = "" 		-- Restriction by unit name ("", "spotter", etc.), not case sensitive
```
Raffson: We should probably refactor this so that these settings can be set through Retribution. I can handle that part if you're unfamiliar with the technicalities... 😅

resources/plugins/plugins.json
```diff
@@ -1,6 +1,7 @@
 [    
     "base",
     "ctld",
+	"Cartsens's Arty",
```
Raffson: I believe this needs to correspond with the mnemonic defined plugin.json

resources/plugins/arty/plugin.json
```diff
@@ -0,0 +1,12 @@
+{
+    "nameInUI": "arty",
```
Raffson: Perhaps something like "Carsten's ArtySpotter" is better suited?

Raffson:
Probably a special thanks in the ReadMe is also appropriate, wdyt?

Turbolious:
> Probably a special thanks in the ReadMe is also appropriate, wdyt?

Yea, that sounds good to me

Raffson:
Superseded by #357

PR #351, opened 2024-06-29
Mirage III mod version 2.5.7.01 support

MetalStormGhost:
Implemented support for the Mirage III mod version 2.5.7.01.


rhoadsy123:
Would we be able to get this mod finished and implemented in Retribution?

Raffson (approved):

PR #358, opened 2024-07-02
Arty plugin Single player version

Turbolious:
Single player version added and called out in Gui

image

image

image

image

image

image


resources/plugins/arty2/CG_ArtySpotter_V1_0_3.lua
```diff
@@ -0,0 +1,538 @@
...
+
+local artyCall     = 0 					-- pilot called arty (from F10 menu)
+local artyRadius   = user_spread		-- Artillery Radius
+local adjustRadius = 20  				-- fire adjustment
+local quantity     = 1   				-- Rounds expanded 
+local quantity_effect = user_quantity	-- Rounds expanded during fire for effect
+local tntEquivalent = 15				-- TNT equivalent for explosion
+local fireDelay = user_fireDelay		-- delay til artillery fires in seconds
+
+local firstShotFired = false
```
Raffson, line 27: The config isn't applied yet, you need to create a global dictionary in this file that corresponds with the names in arty-config.lua

resources/plugins/arty2/plugin.json
```diff
@@ -0,0 +1,38 @@
...
+            "nameInUI": "Max spotting distance. In kilometers.", 
+            "mnemonic": "user_spottingDistance",
+            "defaultValue": 15
+        }
+    ],
+    "scriptsWorkOrders": [
+        {
+            "file": "CG_ArtySpotter_V1_2a_MP.lua",
+            "mnemonic": "arty"
+        }
```
Raffson, line 30: Needs to be adjusted, i.e. the file-name should corresponds with the SP version and the mnemonic would be arty2 in this case.

resources/plugins/arty2/plugin.json
```diff
@@ -0,0 +1,38 @@
...
+    ],
+    "configurationWorkOrders": [
+        {
+            "file": "arty-config.lua",
+            "mnemonic": "arty-config"
```
Raffson, line 35: I'm not entirely sure we can re-use the same name for the arty-config mnemonic, do you mind double checking if this doesn't cause any issues? 😅

Raffson (changes requested):

Raffson:
MP-version should also work for SP-missions, making this PR redundant afaik...

PR #410, opened 2024-10-29
Add Mbot's Call Artillery Script plugin

cedriclmenard:
# Context
While Carsten's artillery script is a great addition, I personally prefer [Mbot's artillery script](https://forum.dcs.world/topic/310506-call-artillery-script/), which uses on-map artillery units and allows for grid coordinates entry through the F-10 radio menu. See [this Youtube video from Mbot](https://www.youtube.com/watch?v=AFPk_RBJjMs) for a nice demo of the script.

To me, this feels more interactive and ties in well with Retribution's Dynamic Campaign and artillery units management.

# Script Requirements
The script calling the `AddFS('GroupName')` LUA function on each artillery group and a `AddFO('UnitName')` for each forward observer (client units able to call artillery). It also includes a small *beep* sound file to play when receiving and sending messages between the forward observers and the fire support groups.

Therefore, the following information needed to be available through the `dcsRetribution` LUA data sent to the plugin scripts:
- The available ground and ship artillery unit DCS group names (for the `AddFS` function), generated from:
    - The available frontline unit DCS group names and types (all artillery units)
    - The available theater object unit DCS group names and types (all artillery units and *Frigate*, *Destroyer* and *Cruiser* ship types)
- The client units DCS unit names for flights with `Armed Recon` role.

# Change Description
- Exposed frontline units groups (both player and enemy in the MissionData dataclass)
- Added `artilleryGroups` and `forwardObserverUnits` data structures to `dcsRetribution` for access in plugin scripts
- Added `inject_other_resource_files` methods to `LuaPlugin` class and related logic. Allows adding resource files to generated miz (used to add the "beep" sound file)

# Possible Improvements/TODOs
- Injected other resource files name deconfliction
- Ensuring ships added to fire support groups (artillery) really do have cannons or rocket artillery available.


Raffson:
Seems like black needs to be updated, and there's also a few mypy errors, but otherwise this looks quite good at first sight 👍

Recommend merging from upstream and updating packages with pip locally, after which you should be able to run black to fix the issues, and similarly for mypy 😉

game/missiongenerator/luagenerator.py
```diff
@@ -210,6 +211,49 @@ def generate_plugin_data(self) -> None:
...
+                ground_artillery_group.add_key_value("groupName", frontline_group.group_name)
+            
+
+        # Add forward observer (FO) (TODO: maybe adding new flight type "Foward Observer"?)
+        forward_observer_object = lua_data.add_item("forwardObserverUnits")
+        for flight in self.mission_data.flights:
+            if not flight.friendly or not len(flight.client_units) > 0:
+                continue
+            if flight.flight_type == FlightType.ARMED_RECON:
+                for unit in flight.client_units:
```
Raffson: We could filter for units with Player or Client skill-levels in order to target the correct units

game/missiongenerator/luagenerator.py
```diff
@@ -210,6 +211,49 @@ def generate_plugin_data(self) -> None:
...
+        # Add artillery that are frontline groups
+        for frontline_group in self.mission_data.player_frontline_groups:
+            if frontline_group.unit_type.unit_class is None:
+                continue
+            if frontline_group.unit_type.unit_class == UnitClass.ARTILLERY:
+                ground_artillery_group = ground_artillery_group_collection.add_item()
+                ground_artillery_group.add_key_value("groupName", frontline_group.group_name)
+            
+
+        # Add forward observer (FO) (TODO: maybe adding new flight type "Foward Observer"?)
```
Raffson, line 257: I tend to agree, but for the time being it might be better to add every client slot as a forward observer?

game/missiongenerator/luagenerator.py
```diff
@@ -210,6 +211,49 @@ def generate_plugin_data(self) -> None:
...
+        forward_observer_object = lua_data.add_item("forwardObserverUnits")
+        for flight in self.mission_data.flights:
+            if not flight.friendly or not len(flight.client_units) > 0:
```
Raffson: What about client slots on OPFOR's side?

game/missiongenerator/luagenerator.py
```diff
@@ -210,6 +211,49 @@ def generate_plugin_data(self) -> None:
...
+                    if unit.unit_type.unit_class == UnitClass.ARTILLERY:
+                        ground_artillery_group = ground_artillery_group_collection.add_item()
+                        ground_artillery_group.add_key_value("groupName", group.group_name)
+                    elif unit.unit_type.unit_class in (UnitClass.CRUISER, UnitClass.DESTROYER, UnitClass.FRIGATE):
+                        # TODO: we assume that these ship classes have guns... Which might not be the case.
+                        ship_artillery_group = ship_artillery_group_collection.add_item()
+                        ship_artillery_group.add_key_value("groupName", group.group_name)
+        
+        # Add artillery that are frontline groups
+        for frontline_group in self.mission_data.player_frontline_groups:
```
Raffson: What about `enemy_frontline_groups`?

game/missiongenerator/luagenerator.py
```diff
@@ -210,6 +211,49 @@ def generate_plugin_data(self) -> None:
...
+                for group in ground_object.groups:
+                    # Check if first unit in group is ground-based or ship artillery
+                    unit = group.units[0]
+                    if unit.unit_type is None:
+                        continue
+                    if unit.unit_type.unit_class == UnitClass.ARTILLERY:
+                        ground_artillery_group = ground_artillery_group_collection.add_item()
+                        ground_artillery_group.add_key_value("groupName", group.group_name)
+                    elif unit.unit_type.unit_class in (UnitClass.CRUISER, UnitClass.DESTROYER, UnitClass.FRIGATE):
+                        # TODO: we assume that these ship classes have guns... Which might not be the case.
```
Raffson: Perhaps we could define a JSON-list or something similar to make the arty-capable ships configurable?

game/missiongenerator/luagenerator.py
```diff
@@ -210,6 +211,49 @@ def generate_plugin_data(self) -> None:
...
             for role, connections in node.connections.items():
                 iads_element.add_data_array(role, connections)
+        
+        # Add artillery and support units info
+        artillery_object = lua_data.add_item("artilleryGroups")
+        ground_artillery_group_collection = artillery_object.get_or_create_item("groundArtillery")
+        ship_artillery_group_collection = artillery_object.get_or_create_item("shipArtillery")
+
+        # First add all artillery units that are theater objects (mostly ships)
+        for cp in self.game.theater.control_points_for(player=True):
```
Raffson: What about OPFOR? 😅

Raffson (changes requested):
Also, a reference to MBot's script in the ReadMe is probably desirable 😅

cedriclmenard:
> Seems like black needs to be updated, and there's also a few mypy errors, but otherwise this looks quite good at first sight 👍
> 
> Recommend merging from upstream and updating packages with pip locally, after which you should be able to run black to fix the issues, and similarly for mypy 😉

Thanks! I've just done that in my latest commit! :)

cedriclmenard:
> Also, a reference to MBot's script in the ReadMe is probably desirable 😅

Added! Thanks!

game/missiongenerator/luagenerator.py
cedriclmenard: I wasn't sure if that was necessary since it is using the `client_units` property, but I'm absolutely not familiar with the codebase, so I've also added a check for player or client skill, thanks! 😄

game/missiongenerator/luagenerator.py
cedriclmenard: I like the idea of having a "dedicated" role to that, plus it can help keep it manageable in large MP scenarios, because I haven't tested the limits of MBot's script. It could be a user toggleable option in the plugins options, but that would necessitate a bit more LUA code to do that.

I can either take more time to update this PR with that or I can prepare another PR with that update, whatever you prefer!

game/missiongenerator/luagenerator.py
cedriclmenard: I honestly didn't take the time before to ensure MBot's script dealt with multiple coalitions, but from reading the code and some testing, it indeed seems to be the case. I've made some changes to consider OPFOR artillery as well! Thanks! 😄

game/missiongenerator/luagenerator.py
cedriclmenard: I agree, but I'm not sure where this would fit in the codebase. It would also be a bit of a research undertaking to do so, plus MBot's scripts already checks the group's ammo and caliber, and only considers rockets and calibers above 70 mm IIRC as possible artillery. I think this would probably rule out the vast majority of cases where ships without guns are added to the artillery groups.

Still, I do think in the long run, having this information somewhere in a JSON would be better. In the meantime, I could probably rename `shipArtillery` to `shipsWithOffensiveArmament` or something like. What do you think?

Raffson:
> > Also, a reference to MBot's script in the ReadMe is probably desirable 😅
> 
> Added! Thanks!

Last but not least, a changelog entry 😅

resources/plugins/artymbot/artymbot-config.lua
```diff
@@ -0,0 +1,27 @@
...
+				AddFS(data.groupName)
+			end
+			if dcsRetribution.plugins.artymbot.shipArtilleryEnable then
+				for _, data in pairs(dcsRetribution.artilleryGroups.shipArtillery) do
+					AddFS(data.groupName)
+				end
+			end
+			for _, data in pairs(dcsRetribution.forwardObserverUnits) do
+				AddFO(data.unitName)
+			end
```
Raffson: Somehow this block decided to use tabs instead of spaces 🤔

game/missiongenerator/luagenerator.py
Raffson: In that case I'd say let's go ahead and merge this already and introduce a separate flight-type later, because I can imagine there'll be a few quirks to this one since it's probably best to keep AI from using that flight-type, since they won't be of much use 😅

game/missiongenerator/luagenerator.py
Raffson: Right... I got confused there as well, the check on skill-level can be skipped because that's exactly what `client_units` will do down the line... my bad 😅

game/missiongenerator/luagenerator.py
Raffson: I don't think renaming to `shipsWithOffensiveArmament` will benefit us a lot, but as you say the checks basically happen in the script. I mainly want to avoid registering handlers that're technically unnecessary, since those will waste CPU time while flying. Nevertheless, I think the performance loss is probably going to be minimal, so I guess we can go ahead with this for the time being 👍

resources/plugins/artymbot/artymbot-config.lua
cedriclmenard: Woah good catch, yeah that's weird! Thanks!

cedriclmenard:
> > > Also, a reference to MBot's script in the ReadMe is probably desirable 😅
> > 
> > 
> > Added! Thanks!
> 
> Last but not least, a changelog entry 😅

Oh yeah you're right! I've added it! Thanks! 😄

Raffson (approved):

Raffson:
Auto-merge enabled 👍 
Btw, are you on our discord? Because there's a Contributor role waiting for you there 😉

cedriclmenard:
Thanks a lot for your help and feedback! I'll join the Discord! 😄

PR #425, opened 2024-12-11
Add TOT generation for all waypoints

Druss99:
This PR aims to add TOT generation for all waypoints. If there is a cleaner way to do this please let me know on discord or via comment.

### DIFF
--- a/qt_ui/windows/mission/flight/waypoints/QFlightWaypointList.py
+++ b/qt_ui/windows/mission/flight/waypoints/QFlightWaypointList.py
@@ -63,8 +63,10 @@ class QFlightWaypointList(QTableView):
             self.model.setHorizontalHeaderLabels(HEADER_LABELS)
 
             waypoints = self.flight.flight_plan.waypoints
+            previous_waypoint = None
             for row, waypoint in enumerate(waypoints):
-                self._add_waypoint_row(row, self.flight, waypoint)
+                self._add_waypoint_row(row, self.flight, previous_waypoint, waypoint)
+                previous_waypoint = waypoint
             self.selectionModel().setCurrentIndex(
                 self.model.index(current_index, 0),
                 QItemSelectionModel.SelectionFlag.Select,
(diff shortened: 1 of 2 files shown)


qt_ui/windows/mission/flight/waypoints/QFlightWaypointList.py
```diff
@@ -63,8 +63,10 @@ def update_list(self) -> None:
             self.model.setHorizontalHeaderLabels(HEADER_LABELS)
 
             waypoints = self.flight.flight_plan.waypoints
+            previous_waypoint = None
```
Raffson: I'd prefer getting rid of this one, and thus the added parameters in the methods as well, and instead cache it like you did with `self.last_tot`, though I would mark them as private with an underscore as prefix, i.e. `self._last_tot`

Other than that it looks pretty good 👍

Raffson (changes requested):

qt_ui/windows/mission/flight/waypoints/QFlightWaypointList.py
Druss99: Should be cached now!

Druss99:
Latest commit fixes the caching to be a bit cleaner

Druss99:
Still need to add logic to do the kneeboards

qt_ui/windows/mission/flight/waypoints/QFlightWaypointList.py
```diff
@@ -112,18 +118,34 @@ def on_changed(self) -> None:
...
         prefix = ""
         time = flight.flight_plan.tot_for_waypoint(waypoint)
         if time is None:
             prefix = "Depart "
             time = flight.flight_plan.depart_time_for_waypoint(waypoint)
         if time is None:
-            return ""
+            prefix = ""
+            timedelta = flight.flight_plan.travel_time_between_waypoints(
+                self._last_waypoint, waypoint
```
Raffson, line 138: The only concern I have at this point is that `self._last_waypoint` may be uninitialized, though theoretically I can't think of a case where this would actually present itself... 🤔

Raffson (approved):

PR #429, opened 2024-12-17
Recovery tanker support

Druss99:
This PR adds the functionality to create a recovery tanker flight. I expect at least a few bugs as well as suggested code changes. So far things look decent from my testing.


Druss99:
Everything seems to be good from a retribution standpoint though I'm having issues getting the tanker to respond in game, more troubleshooting needed

game/missiongenerator/aircraft/waypoints/recoverytanker.py
```diff
@@ -0,0 +1,85 @@
...
+        for key, value in theater_objects.items():
+            # Check name and position in case there are multiple of same carrier.
+            if name in key and value.theater_unit.position == carrier_position:
+                return value.dcs_group_id
+        raise RuntimeError(
+            f"Could not find a carrier in the mission matching {name} at "
+            f"({carrier_position.x}, {carrier_position.y})"
+        )
+
+    def configure_tanker_tacan(self, waypoint: MovingPoint) -> None:
```
Raffson: I believe this method is redundant since tacan should be set up as part of the flightconfigurator 🤔

Druss99:
Removed dependency on Unit_Map, validated tanker does allow you to refuel. Ready for review

game/missiongenerator/missiondata.py
```diff
@@ -58,6 +59,8 @@ class CarrierInfo(UnitInfo):
     tacan: TacanChannel
     icls_channel: int | None
     link4_freq: RadioFrequency | None
+    position: Point
+    group_id: int
```
Raffson: We could just reference the group itself, which should contain both of these properties and would allow us to reference its waypoints if required without too much hassle, does that make sense? 😅

game/missiongenerator/aircraft/waypoints/recoverytanker.py
```diff
@@ -0,0 +1,88 @@
...
+            ):
+                waypoint.tasks[0] = SetUnlimitedFuelCommand(False)
+            else:
+                waypoint.tasks.insert(0, SetUnlimitedFuelCommand(False))
+
+        clouds = self.flight.squadron.coalition.game.conditions.weather.clouds
+        waypoint.add_task(Tanker())
+        group_id = self._get_carrier_group_id()
+        speed = knots(250).meters_per_second
+        altitude = feet(6000).meters
```
Raffson: Keep in mind that this happens during mission generation time, so this means you won't see the actual altitude which will be used during generation in the waypoints tab. Ideally we initialize this during creation of the FlightWaypoint, although it's not _that_ big of a deal 😅

game/commander/objectivefinder.py
```diff
@@ -228,13 +228,16 @@ def farthest_friendly_control_point(self) -> ControlPoint:
...
+    def closest_friendly_control_point(self) -> ControlPoint | None:
         """Finds the friendly control point that is closest to any threats."""
         threat_zones = self.game.threat_zone_for(not self.is_player)
 
         closest = None
         min_distance = meters(math.inf)
         for cp in self.friendly_control_points():
+            # If statement for https://github.com/dcs-liberation/dcs_liberation/issues/2693
+            if cp.is_fleet:
+                return closest
```
Raffson: I'm not entirely sure I understand the reasons behind this one 😅

game/ato/flightplans/shiprecoverytanker.py
```diff
@@ -0,0 +1,51 @@
...
+    def patrol_duration(self) -> timedelta:
+        return self.flight.coalition.game.settings.desired_tanker_on_station_time
+
+
+class Builder(IBuilder[RecoveryTankerFlightPlan, PatrollingLayout]):
+    def layout(self) -> PatrollingLayout:
+
+        builder = WaypointBuilder(self.flight)
+        # TODO: Propagate the ship position to the Tanker's TOT,
+        # so that we minimize the tanker's need to catch up to the carrier.
```
Raffson: I'm not entirely sure that we have access to it, but if we could reference the wind direction here that would help us fine-tune the position of the waypoint. Will try to check what possibilities we have tomorrow 🤞

Raffson (changes requested):
There's one fundamental thing missing I believe, i.e. a way to specifically assign the role to specific aircraft, because I don't think the pilots of the Hornets & Tomcats will be happy when a KC-135 shows up... 😅

game/commander/objectivefinder.py
Druss99: So the autoplanner automatically tries to plan a theater tanker if that option is selected. Since we are reusing the refueling flight plan as of right now it could try to plan a recovery tanker if that control point was the closest. If the carrier was sunk on the previous turn it would fail to generate a mission because it can't generate a recovery tanker around a sunken carrier. This change would essentially make it never plan a recovery tanker as part of the "always plan theater refueler"

Druss99:
> There's one fundamental thing missing I believe, i.e. a way to specifically assign the role to specific aircraft, because I don't think the pilots of the Hornets & Tomcats will be happy when a KC-135 shows up... 😅

Technically the autoplanner would never plan a recovery tanker flight because of the changes I made to closest_friendly_cp. That being said I think I am in agreement that creating a separate flight task called "Recovery" and completely separating it from "Refueling" is probably better. That way we can restrict what aircraft can do the task and we can probably implement an option for autoplanning recovery tanker tasks.

game/commander/objectivefinder.py
Druss99: If we separate this into its own flight task these changes won't be needed which is what I think I am going to do

game/missiongenerator/aircraft/waypoints/recoverytanker.py
Druss99: I think it is needed since it looks like refueling tasks get it through: https://github.com/dcs-retribution/dcs-retribution/blob/dev/game/missiongenerator/aircraft/waypoints/racetrack.py

Druss99: See line 78

game/missiongenerator/aircraft/waypoints/recoverytanker.py
Druss99: I see where to add it to the waypoints tab but then does the altitude set here get overwritten or?

Druss99:
Still need to add an option to do autoplanning for recovery tankers

game/missiongenerator/aircraft/waypoints/recoverytanker.py
Raffson: Right my bad, it's the waypoint-action to actually activate the TCN, but since the logic looks quite similar, I was thinking we could maybe refactor it?

game/missiongenerator/aircraft/waypoints/recoverytanker.py
Raffson: The other way around, i.e. whatever you would set in the waypoints tab would get overwritten by this part. Does that make sense? 😅

Raffson:
> Still need to add an option to do autoplanning for recovery tankers

A few ways we can handle this, though I'm still not entirely sure what the neatest way would be... for now I would say let's leave it out since we'll probably still have to iron out a bug or two during testing 😅

game/ato/flightplans/shiprecoverytanker.py
Raffson: We have the wind-direction at our disposal, but we'll also need to TOT for an estimated location of the carrier. However, the TOT is currently a bit of a mess because it doesn't take into account when other flight start returning to the carrier, so that'd be the first thing to fix on my end for the auto-planner to work decently... yes, despite what I initially said about holding off on that I'm actually looking into it 😅

Raffson (approved):

Raffson (approved):

PR #432, opened 2024-12-20
bugfix if settings directory doesnt exist

Druss99:
If the settings directory doesn't exist it would fail to create Default.zip, now the directory is always created


qt_ui/windows/settings/QSettingsWindow.py
```diff
@@ -575,6 +575,8 @@ def save_settings(self):
 
     def load_default_settings(self):
         sd = settings_dir()
+        if not sd.exists():
+            sd.mkdir()
```
Raffson: To prevent future issues regarding this topic, perhaps we should move this logic into the settings_dir function? 🤔

Raffson (approved):

PR #434, opened 2024-12-29
Faction editor update

Druss99:
This PR allows users to edit a faction and save it as a new faction. The two improvements I am unsure how to implement is:
- Currently when a checkbox is unticked, adding a unit will retick the checkbox, probably need a way to run through the checkboxes, see what is checked, and recreate them. Or just change it to not recreate all check boxes and instead just add a checkbox for the unit added
- Little python boxes pop up (presumably due to the _format method) when the new game wizard is launched

### DIFF
--- a/qt_ui/windows/newgame/WizardPages/QFactionSelection.py
+++ b/qt_ui/windows/newgame/WizardPages/QFactionSelection.py
@@ -23,6 +23,7 @@ from PySide6.QtWidgets import (
     QFileDialog,
 )
 
+from game import persistency
 from game.armedforces.forcegroup import ForceGroup
 from game.ato import FlightType
 from game.campaignloader import Campaign
@@ -554,11 +555,15 @@ class QFactionSaver(QDialog):
         self.setLayout(layout)
 
     def save_faction(self) -> None:
+        # TODO: Reload factions combobox on save
         self.faction.name = self.name_text.text()
         self.faction.description = f"<p>{self.description_text.text()}</p>"
         self.faction.authors = self.authors_text.text()
+        user_faction_path = persistency.factions_dir()
 
-        fd = QFileDialog(caption="Save Faction", filter="*.json")
+        fd = QFileDialog(
+            caption="Save Faction", directory=str(user_faction_path), filter="*.json"
+        )
         fd.setAcceptMode(QFileDialog.AcceptMode.AcceptSave)
         if fd.exec_():
             json_filename = fd.selectedFiles()[0]
--- a/qt_ui/windows/newgame/WizardPages/QFactionSelection.py
+++ b/qt_ui/windows/newgame/WizardPages/QFactionSelection.py
@@ -55,12 +55,13 @@ class QFactionUnits(QScrollArea):
         combo_layout: Optional[QHBoxLayout] = None,
     ) -> int:
         counter += 1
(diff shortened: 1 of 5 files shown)


qt_ui/windows/newgame/WizardPages/QFactionSelection.py
```diff
@@ -45,12 +55,13 @@ def _add_checkboxes(
         combo_layout: Optional[QHBoxLayout] = None,
     ) -> int:
         counter += 1
-        for i, v in enumerate(sorted(units, key=lambda x: str(x)), counter):
-            cb = QCheckBox(str(v))
-            cb.setCheckState(Qt.CheckState.Checked)
-            self.checkboxes[str(v)] = cb
-            grid.addWidget(cb, i, 1)
-            counter += 1
+        if len(units) > 0:
```
Raffson: Any particular reason for this if-statement? Doesn't look like it provides any added value 🤔

qt_ui/windows/newgame/WizardPages/QFactionSelection.py
```diff
@@ -228,6 +321,22 @@ def updateFactionUnits(self, units: Union[set, list]):
...
+        hbox = QHBoxLayout()
+        hbox.addWidget(cb)
+        hbox.addWidget(add_button)
+        if cb.count() == 0:
+            cb.hide()
+            add_button.hide()
+        else:
+            cb.show()
+            add_button.show()
+        return hbox
```
Raffson, line 337: These are causing the popups, changing them to `cb.setEnabled(False)` or `cb.setEnabled(True)` (and for `add_button` respectively) should fix the issue. If you really want to hide them, you'll need to use `setStyleSheet` and use CSS instead ;)

qt_ui/windows/newgame/WizardPages/QFactionSelection.py
```diff
@@ -284,6 +393,17 @@ def __init__(self, parent=None):
             if r == "USA 2005":
                 self.blueFactionSelect.setCurrentIndex(i)
 
+        self.saveBlueFactionButton = QtWidgets.QPushButton("Save as new faction")
+        self.saveRedFactionButton = QtWidgets.QPushButton("Save as new faction")
```
Raffson: No need to mention `QtWidgets`, `QPushButton` is already imported...

qt_ui/windows/newgame/WizardPages/QFactionSelection.py
```diff
@@ -380,10 +500,68 @@ def _filter_selected_units(qfu: QFactionUnits) -> Faction:
         qfu.updateFactionUnits(fac.missiles)
         return fac
 
+    def show_save_faction_dialog(self, faction: Faction):
+        dialog = QFactionSaver(faction)
+        dialog.exec_()
```
Raffson: Prefer `exec` instead, `exec_` is an old thing because `exec` is also a Python function...

Raffson (changes requested):

Druss99:
All errors fixed, will potentially look to refactor some of this longterm but I'd rather ship in 1.4 and afterwards see if the juice is worth the squeeze on a refactor

Raffson (approved):

PR #437, opened 2024-12-30
Fix airwing config not properly saving or loading

Druss99:
Previously, saving an airwing config would not actually save your exact air wing because the tabs were not being properly updated. You would end up with whatever squadrons you added but any that were removed would not be removed when the air wing was saved. That is now fixed. Additionally, loading an airwing config after starting a new campaign would throw errors due to carrier names not matching up, this is now fixed.

### DIFF
--- a/game/campaignloader/campaignairwingconfig.py
+++ b/game/campaignloader/campaignairwingconfig.py
@@ -71,9 +71,25 @@ class CampaignAirWingConfig:
         cls, data: dict[Union[str, int], Any], theater: ConflictTheater
     ) -> CampaignAirWingConfig:
         by_location: dict[ControlPoint, list[SquadronConfig]] = defaultdict(list)
+        carriers = theater.find_carriers()
+        lhas = theater.find_lhas()
         for base_id, squadron_configs in data.items():
             if isinstance(base_id, int):
                 base = theater.find_control_point_by_airport_id(base_id)
+            elif "CVN" in base_id:
+                if carriers:
+                    base = carriers.pop(0)
+                else:
+                    raise ValueError(
+                        f"Air wing config incompatible due to lack of carriers in theater"
+                    )
+            elif "LHA" in base_id:
+                if lhas:
+                    base = lhas.pop(0)
+                else:
+                    raise ValueError(
+                        f"Air wing config incompatible due to lack of LHAs in theater"
+                    )
             else:
                 base = theater.control_point_named(base_id)
 
--- a/game/campaignloader/campaignairwingconfig.py
+++ b/game/campaignloader/campaignairwingconfig.py
@@ -4,8 +4,6 @@ from collections import defaultdict
 from dataclasses import dataclass
 from typing import Any, Optional, TYPE_CHECKING, Union
 
-from game.dcs.shipunittype import ShipUnitType
-
 from game.ato.flighttype import FlightType
 from game.theater.controlpoint import ControlPoint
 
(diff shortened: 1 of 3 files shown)


qt_ui/windows/AirWingConfigurationDialog.py
```diff
@@ -823,6 +823,8 @@ def save_config(self) -> None:
         )
         fd.setAcceptMode(QFileDialog.AcceptMode.AcceptSave)
         if fd.exec_():
+            for tab in self.tabs:
+                tab.apply()
```
Raffson, line 838: Could you elaborate this change? It might just be me, but I can't seem to understand how this applies to the rest of the code 😅

game/campaignloader/campaignairwingconfig.py
```diff
@@ -71,9 +71,25 @@ def from_campaign_data(
...
     ) -> CampaignAirWingConfig:
         by_location: dict[ControlPoint, list[SquadronConfig]] = defaultdict(list)
+        carriers = theater.find_carriers()
+        lhas = theater.find_lhas()
         for base_id, squadron_configs in data.items():
             if isinstance(base_id, int):
                 base = theater.find_control_point_by_airport_id(base_id)
+            elif "CVN" in base_id:
+                if carriers:
+                    base = carriers.pop(0)
```
Raffson: Is it possible that we assign a squadron to an enemy carrier here? 🤔

qt_ui/windows/AirWingConfigurationDialog.py
Druss99: Currently we don't have a way to access the updated squadron data, we do that intentionally so that a user can revert their changes if they want. Because of that, when you save it would save squadrons even if they were deleted in the UI. To fix it I just called the apply method so the squadron gets updated when a user saves the faction. I also have since added a dialog box to warn the user that they can't revert changes after saving a squadron.

game/campaignloader/campaignairwingconfig.py
Druss99: Yes, this is now fixed, the code is a little messy but hopefully not too messy for your taste haha

qt_ui/windows/AirWingConfigurationDialog.py
```diff
@@ -836,7 +849,7 @@ def _build_air_wing(self) -> dict:
             for s in sqs:
                 cp = s.location.at
                 if isinstance(cp, Point):
-                    key = s.location.name
+                    key = s.location.full_name
```
Druss99, line 852: I'm sure you'll ask about this, I made this change to use the full name so that carriers would save with a very deterministic name. The only potential issue with this change is that loading squadrons from a campaign with different types of carriers (ex. a faction that has a kuz and a cvn) it would fail to load. While that is less than ideal, the current implementation wouldn't let you load a air wing config at all unless it was the same exact campaign with the same exact carrier names so overall I see it as an improvement with perhaps a better solution out there that I can't come up with myself but I will defer to your judgement on that one

game/campaignloader/campaignairwingconfig.py
```diff
@@ -71,11 +71,27 @@ def from_campaign_data(
...
+                        base = next((c for c in carriers if not c.captured), None)
+                    elif base_id == "Blue CV":
+                        base = next((c for c in carriers if c.captured), None)
+                    elif base_id == "Red LHA":
+                        base = next((l for l in lhas if not l.captured), None)
+                    elif base_id == "Blue LHA":
+                        base = next((l for l in lhas if l.captured), None)
+
+            if base is None:
+                raise ValueError(f"No valid control points found {base_id}")
```
Raffson: Perhaps log an error and skip the base? Raising an exception seems a little destructive since it would mean the air-wing won't get loaded up at all...

Raffson (approved):

PR #443, opened 2025-01-03
Skynet additional settings

Astro-739:
Additional settings for Skynet offer interesting options for campaign building and playing.

This PR contains the following:
1. Separate ActMobile settings for SHORAD and MERAD
2. Settings for GoLiveRange

Separate ActMobile settings for SHORAD and MERAD
SHORAD and MERAD units react differently to the ActMobile commands. SHORAD units relocate and turn off and on their radar  pretty quickly. MERAD units take longer to prepare to relocate and are often in a group that starts in a nice layout, but after relocating they just end up on a road in a row. So, different settings for SHORAD and MERAD units helps.

GoLiveRange setting
To be able to adjust the GoLiveRange of a SAM is a very interesting element for campaign design. This applies mainly to long range SAMs, that now engage a target on the edge of their interceptor range, making it relatively easy to dodge. When the GoLiveRange is set closer to the SAM, the missile is more dangerous and harder to dodge.
Also this gives the opportunity to use the very long range SAMs (e.g. SA-5/20) and reduce their range to not overlap the entire map.


Astro-739:
I have tested these setting in the past weeks with the Into the Hornet's Nest campaign and they are a great addition.

Druss99:
Some people in the discord were talking about how it would be nice to adjust the harm detection chance for all SAMs. This could get messy because some SAMs shouldn't be able to detect harms ever and others have a default of 90ish so one global setting wouldn't handle it but this PR might be a good place to add that stuff if you're interested

Astro-739:
Yes, definitely interested in more skynet settings. But I agree with you, that a generic setting gets too messy. I am on the road at the moment, will have a look into it later today or tomorrow.

resources/plugins/skynetiads/skynetiads-config.lua
```diff
@@ -18,10 +18,21 @@ if dcsRetribution and SkynetIADS then
     local includeBlueInRadio = false
     local debugRED = false
     local debugBLUE = false
-    local actMobile = false
-    local actMobileMaxEmissionTime = 120
-    local actMobileMinimumScootDistance = 500
-    local actMobileMaximumScootDistance = 3000
+    local actMobile = false    -- did not change to actMobile_shorad for backwards compatibility
```
Raffson, line 21: Backwards compatibility is technically irrelevant in this case, as long as the plugin's options don't change (i.e. the mnemonic in plugin.json) there shouldn't be an issue. Just an FYI, no need to change it since the comment clears up the confusion 👍

Raffson (commented):
I don't see any major issues at this point, so I'd say we can merge this already if everyone agrees? 😅

Druss99:
sounds good to me

Raffson (approved):

PR #450, opened 2025-01-10
Autocreate package fix

Druss99:
Bugfix WRT: https://discord.com/channels/1015931619187621999/1025759185511653387/1326666368799735808

Looks like this bug would occur any time you attempt to use the auto create feature and an air assault (or other primary task flight) was a helicopter. Helo would be set to true and all of the escorts, sead sweep, etc would fail to find a squadron because they are looking for only helo squadrons. Maybe this was preferred in the past to keep a flight package of only helos if a helo task is the primary flight? Unsure but this fixes the bug.

Fix: Use plan.perferred_type.helicopter instead of heli/primaryflight.is_helo

Please review, it's only a few lines changed but the affects might be farther reaching than I understand. I presume we were using the primary flight is helo attribute for some reason but I can't see why it would be better than plan.preferred_type.helicopter

### DIFF
--- a/game/commander/packagebuilder.py
+++ b/game/commander/packagebuilder.py
@@ -48,6 +48,7 @@ class PackageBuilder:
         using release_planned_aircraft.
         """
         target = self.package.target
+        heli = False
         pf = self.package.primary_flight
         if pf:
             target = (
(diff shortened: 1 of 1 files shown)


Druss99:
Intended behavior that needs better error handling

Raffson:
> Intended behavior that needs better error handling

Or restrictions in the selectors of the auto-create package dialog 😁

PR #456, opened 2025-01-17
Neutral Base support

Druss99:
This PR implements support for Neutral Bases in retribution campaigns.
TODO:

Update pydcs to support (currently I have my local pydcs edited to get this to function for testing) Below is my current blob I'm using to test:
`
def set_blue(self):
self.set_coalition("BLUE")

def set_red(self):
self.set_coalition("RED")

def set_neutral(self):
self.set_coalition("NEUTRAL")

def set_coalition(self, side):
if side.lower() in ["red", "blue", "neutral"]:
self.coalition = side.upper()
return True
return False

def is_red(self):
return self.coalition == "RED"

def is_blue(self):
return self.coalition == "BLUE"

def is_neutral(self):
if self.dynamic_spawn:
return self.coalition == "NEUTRAL"
else:
return False
`

Better handling of neutral unit generation. Right now I set all units at a neutral point to dead except static objects, the objects populate in the mission as bluefor objects since my neutral coalition is just a copy of the blue coalition with player set to None. Ideally if the control point is neutral it would not generate any objects? Though that might be odd for the player if they overfly an airfield with nothing around it and then they capture it and the next turn there are warehouses and stuff everywhere, unsure what the best way to handle this is

Potentially fix the API, I made some changes to allow null values and made the blue boolean value optional in the model.py files, I don't fully understand the impact of those changes so there may be issues there

Use something other than dynamic spawn to designate a CP as neutral

Fix typing that is now horribly broken


game/missiongenerator/missiongenerator.py
```diff
@@ -373,9 +373,15 @@ def generate_warehouses(self) -> None:
...
-                warehouse["coalition"] = (
-                    "blue" if tmu.theater_unit.ground_object.coalition.player else "red"
-                )
+                if tmu.theater_unit.ground_object.coalition.player:
+                    warehouse["coalition"] = "blue"
+                elif (
+                    tmu.theater_unit.ground_object.coalition.neutral
+                    and not tmu.theater_unit.ground_object.coalition.player
+                ):
+                    warehouse["coalition"] = "neutral"
```
Raffson: Shouldn't we first check for `neutral` before checking for `player`?

game/missiongenerator/tgogenerator.py
```diff
@@ -867,7 +867,12 @@ def create_helipad(
             pad.position,
             self.m.terrain,
         ).dict()
-        warehouse["coalition"] = "blue" if self.cp.coalition.player else "red"
+        if self.cp.coalition.player:
+            warehouse["coalition"] = "blue"
+        elif self.cp.coalition.neutral and not self.cp.coalition.player:
+            warehouse["coalition"] = "neutral"
```
Raffson: Same here?

game/missiongenerator/tgogenerator.py
```diff
@@ -994,7 +999,12 @@ def create_ground_spawn_roadbase(
...
```
Raffson: And here... 😅

game/missiongenerator/tgogenerator.py
```diff
@@ -1120,7 +1130,12 @@ def create_ground_spawn_large(
...
```
Raffson: And all other remaining occurrences? 😅

game/missiongenerator/triggergenerator.py
```diff
@@ -242,6 +245,51 @@ def _generate_capture_triggers(
...
+                    inverted_recapture_trigger.add_condition(FlagIsTrue(flag=flag))
+                    script_string = String(
+                        f'base_capture_events[#base_capture_events + 1] = "{cp.id}||{attack_coalition_int}||{cp.full_name}"'
+                    )
+                    inverted_recapture_trigger.add_action(DoScript(script_string))
+                    inverted_recapture_trigger.add_action(ClearFlag(flag=flag))
+                    self.mission.triggerrules.triggers.append(
+                        inverted_recapture_trigger
+                    )
+
```
Raffson, line 283: Do you mind refactoring this into a separate method? 😅 
Perhaps something like `set_neutral_capture_triggers`?
Also, are we sure this doesn't interfere with the exisiting triggers? 🤔

game/theater/controlpoint.py
```diff
@@ -813,7 +826,8 @@ def find_ground_objects_by_obj_name(
                 found.append(g)
         return found
 
-    def is_friendly(self, to_player: bool) -> bool:
+    # TODO: The below two functions need to be refactored for Neutral base support or we need a new function that is called before these two functions to determine if the base is neutral or not.
+    def is_friendly(self, to_player: bool) -> Optional[bool]:
```
Raffson: Return-type seems off here, it's still a bool to be returned, but `to_player` should become `Optional[bool]` if my understanding is correct, and more importantly, _if_ we want to continue with the approach of having player & neutral as split variables 😅

Raffson (changes requested):

Druss99:
I am now using this branch of pydcs to test: https://github.com/Druss99/pydcs/tree/neutralbasesupport

Everything seems to function as intended except for the UI. There are a good number of typscript changes that I am not sure how to fix to include sam threat range not showing and the build failing due to a typescript test failure. I think this is relatively close to being over the line but it definitely needs a little help from @Raffson and vigorous testing

Druss99:
Most of the typescript issues have been fixed, I could potentially minimize the amount of backend that I edited by simply passing blue as a true false into the UI since neutral bases should never have things like threat ranges, neutral flight plans, etc so setting them to be true or false shouldn't affect much but let me know. Currently I am passing in a number value and just comparing to that. This definitely needs large scale review and testing

Druss99:
TODO: Implement neutral FOBs (This can and should probably go in a separate PR)

game/missiongenerator/triggergenerator.py
Druss99: Tested the triggers and they function as intended but refactoring into a separate method is still TODO

Druss99:
Reverted all API changes, instead the python handles the translation to boolean for the UI. We may want to adjust this slightly so that supply routes to neutral bases show in a different color or don't show at all. 

From my understanding, the blue flag really only affects things like displaying sam threat ranges, flights, and supply routes. Since most of those things don't apply to neutral bases it seems fine to keep using a boolean in the API for now. Reference previous commits and comments for context

game/coalition.py
```diff
@@ -95,8 +98,9 @@ def __getstate__(self) -> dict[str, Any]:
         state = self.__dict__.copy()
         # Avoid persisting any volatile types that can be deterministically
         # recomputed on load for the sake of save compatibility.
-        del state["_threat_zone"]
-        del state["_navmesh"]
+        if state["player"] != Player.NEUTRAL:
```
Raffson, line 103: Why would you want to keep the threat-zones & navmesh for neutral exactly? 🤔

game/commander/objectivefinder.py
```diff
@@ -10,12 +10,14 @@
...
     Airfield,
     ControlPoint,
+    Player,
     Fob,
     FrontLine,
     MissionTarget,
     OffMapSpawn,
     ParkingType,
     NavalControlPoint,
+    Player,
```
Raffson, line 20: Double import 😅

game/commander/objectivefinder.py
```diff
@@ -35,7 +37,7 @@
 class ObjectiveFinder:
     """Identifies potential objectives for the mission planner."""
 
-    def __init__(self, game: Game, is_player: bool) -> None:
+    def __init__(self, game: Game, is_player: Player) -> None:
         self.game = game
         self.is_player = is_player
```
Raffson, line 42: Perhaps we should rename this to `self.player` create some method like `get_opposing_player` or something similar to handle the "inversions" below? I'd even consider defining this method on the `Player` enum itself... Wdyt?

requirements.txt
```diff
@@ -33,7 +33,7 @@ pluggy==1.5.0
 pre-commit==4.2.0
 pydantic==2.11.0b2
 pydantic-settings==2.8.1
-pydcs @ git+https://github.com/dcs-retribution/pydcs@4a5e3ee63d3a487841c7b88c0a5bf58811567d6c
+pydcs @ git+https://github.com/Druss99/pydcs@3f44568904fb1ca2460eaded8bc892415c27a227
```
Raffson: I guess we need to merge some things to pydcs and point this link back to the correct place 😅

resources/scripts/MissionScripting.original.lua
```diff
@@ -13,9 +13,9 @@ local function sanitizeModule(name)
 end
 
 do
-	sanitizeModule('os')
-	sanitizeModule('io')
-	sanitizeModule('lfs')
+	--sanitizeModule('os')
+	--sanitizeModule('io')
+	--sanitizeModule('lfs')
```
Raffson: This should be uncommented again

Raffson (commented):
This must've been a pain to replace every instance of `not captured` or `not player` 😅
I'm thinking whether or not we should replace instances of blue/red if <something> else red/blue with more elegant solution right away, or do this gradually over time 🤔

game/commander/objectivefinder.py
Druss99: Taking a second look at this, I think I get what your saying here, maybe we create a player class and create a few handy methods that we can use across retribution to handle the If is_player and inversions of that, etc. Is that what you're getting at? Having all of the checking player stuff nested in some methods in the player class?

Druss99:
Neutral FOB support added. For a neutral fob place a RED KrAZ6322. Also added automatic air assault planning against neutral bases. Save compat looks good but I did not run through a full mission using an old campaign, there is a chance that accepting the state.json for previously generated campaigns is broken but I will test that when I return from vacation.

The only oddity I noticed was the redfor helis did not actually drop troops during the air assault. If this is reproducible I will look into it further.

Druss99:
Fixed a lot of bugs:
1. Redfor not dropping troops on air assault
2. Neutral FOBs not spawning buildings
3. Neutral FOBs spawning blue buildings
4. Cheat capture not working for neutral FOBs

game/coalition.py
Druss99: Fixed, I think I had this in because I thought it was causing issues and it was probably just hiding other real issues

game/missiongenerator/triggergenerator.py
Druss99: Done

Raffson (approved):

PR #464, opened 2025-01-25
Update swedish military assets pack

Druss99:
This pr fixes the currently broken swedish military assets pack with as minimal changes as possible


game/factions/faction.py
```diff
@@ -575,8 +575,10 @@ def apply_mod_settings(self, mod_settings: Optional[ModSettings] = None) -> None
             self.remove_preset("David's Sling (Semicircle)")
         # swedish military assets pack
         if not mod_settings.swedishmilitaryassetspack:
-            self.remove_vehicle("BV410_RBS70")
-            self.remove_vehicle("BV410_RBS90")
```
Raffson, line 579: These seem to have been replaced by RBS_70 & RBS_90, correct? If so, there's some other implications that need to be addressed, such as its usage in the Pretense generator 😅

resources/units/ground_units/CV9040.yaml
```diff
@@ -8,5 +8,4 @@ origin: Sweden
 price: 10
 role: Infantry Fighting Vehicle
 variants:
-  Strf 9040 IFV: {}
-  CV9040 IFV: {}
```
Raffson: Recommend reverting all changes to existing yaml files. The only thing that may need to be adjusted is the filename itself, i.e. to match the pydcs-id of the unit, but the variants should remain unchanged 😬

resources/units/ground_units/Artillerisystem08_SGR77B.yaml
```diff
@@ -0,0 +1,13 @@
...
+  a fully automated 155 mm/L52 gun howitzer and a M151 Protector remote
+  controlled weapon station mounted on a modified 6x6 chassis of the
+  Volvo A30D all-terrain articulated hauler.
+introduced: 2016
+manufacturer: Bofors
+origin: Sweden
+price: 25
+role: Self-Propelled Gun
+variants:
+  "[CH] Archer SPG SGR77B": {}
```
Raffson, line 13: I've noticed you're using `[CH]`, but I'm wondering whether that's really a good idea... On one hand I like it because it immediately identifies the mod, but on the other I'm afraid this name might perhaps change in the future and thus require a change here as well, which should be avoided at all costs because it's a recipe for save-compatibility issues 😅

Raffson (changes requested):

resources/units/ground_units/Artillerisystem08_SGR77B.yaml
Druss99: Since they're only variant names I think leaving it in is fine, if it changes in the future we can just add more variants like I did this time with the units that changed names

Raffson (approved):

Raffson (approved):

PR #471, opened 2025-01-27
fix save compat for changed mod unit names

Druss99:
some mod units were removed or renamed from the CH pack, migration unpickler updated to fix these changes

### DIFF
--- a/resources/units/ground_units/RBS-70.yaml
+++ b/resources/units/ground_units/RBS-70.yaml
@@ -10,4 +10,5 @@ price: 15
 role: Surface-to-Air Missile Launcher
 variants:
   RBS 70 EldE 70 Stationary SAM LN: {}
+  BV 410 RBS 70: {}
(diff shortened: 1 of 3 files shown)


resources/units/ground_units/RBS-70.yaml
```diff
@@ -10,4 +10,5 @@ price: 15
 role: Surface-to-Air Missile Launcher
 variants:
   RBS 70 EldE 70 Stationary SAM LN: {}
+  BV 410 RBS 70: {}
```
Raffson, line 13: Could you test whether the fix works without changes to to unit yamls?

Druss99: I originally only edited the migration unpickler but then it complained about the variant name so I adjusted the unit yamls and it ran fine

Raffson (approved):

PR #475, opened 2025-01-28
Remove SEAD and SEAD Escort mission type from Gripen

Starfire13:
Sweden doesn't use anti-radiation missiles and Gripens don't perform those missions.


Raffson (commented):
Isn't a matter of just adding a SEAD payload? 😅

Raffson:
Nvm, should've read the description first 😂

Raffson (approved):

PR #480, opened 2025-02-05
Mission directory and missile site bugfixes

Druss99:
Fixes errors when mission directory does not exist and bug where mod missile sites don't get removed when the mod is disabled. Also renamed some icon files to be proper and not need manual mappings


Druss99:
Ready for merge

Raffson (approved):

PR #481, opened 2025-02-05
CurrentHill Ukraine Assets Pack

Druss99:
Implements support for Ukraine Asset Pack


Druss99:
Ready for merge

resources/ui/units/aircrafts/icons/CH_Su-27P1M_24.jpg
Raffson, line 1: Did you push icons from other packs the same way? To be honest I haven't given it a closer look but now that I do I'm noticing that this is not consistent with all other icons 😅 
Usually the way to go about this is to use DCS' model-viewer, which allows you to get a perfect side-angle view of the aircraft, and also allows for playing with the sun's position to get a better lighting. A simple screenshot with the correct aspect ratio can then be shrunk to the icon's dimensions.

For now something that I would consider as a blocker to merge, but it'd be nice to have consistency of course 😅

resources/units/aircraft/Su-24MU.yaml
```diff
@@ -0,0 +1,24 @@
...
+introduced: 1983
+manufacturer: Sukhoi
+origin: USSR/Russia
+price: 14
+role: Attack
+max_range: 200
+default_livery: af standard
+variants:
+  "[CH] Su-24MU": {}
+tasks:
```
Raffson, line 18: You could probably add SEAD & SEAD Sweep to this, though I'm a bit doubting about SEAD Escort, though in theory it should be able to so I'd simply give it a lower priority than the rest, say something like 200-ish

resources/ui/units/aircrafts/icons/CH_Su-27P1M_24.jpg
Starfire13: Russia asset pack uses side views, USA doesn't (but that's just one aircraft; B-21). Sweden pack uses sideview for the chopper and but not for Gripen. China asset pack has no aircraft.

resources/units/aircraft/Su-24MU.yaml
Starfire13: It has no TALDs or anti-radiation missiles though. The Kh-25 it carries is the ML, not the MP variant. That's laser guided, not an ARM. The Kh-29 variant it carries is a TV-giuded missile.

resources/ui/units/aircrafts/icons/CH_Su-27P1M_24.jpg
Raffson: Thanks for identifying the ones I missed 😅
Will resolve and merge for now, that way the new campaign can be merged as well.

Raffson (approved):

PR #484, opened 2025-02-08
Update F/A-18E/F/G Super Hornet Mod to 2.4.0

Drexyl:
Update support for CJS Super Hornet Mod to 2.4.0.

Combines Superbug Tanker AI Mod.

Remove references to old Superbug Tanker Mod.

Added missing F/A-18F Tanker to Blufor Current.


pydcs_extensions/fa18efg/fa18efg.py
Raffson, line 1: There's still some work to be done here because there's too many weapons here. I can also see custom weapons from the mod being referenced from the `Weapons` class, which should not be the case. Instead those custom weapons need to be extracted from the pydcs export and included in the pydcs-extension, where the weapons should get injected into the pydcs layer. Let me know if you need help with this 😉

resources/factions/blufor_current.json
Raffson, line 1: I believe spaces got replaced with tabs here, which we'd like to avoid because it creates too many changes for git, which in turn will eventually result into a larger git repo than necessary, does that make sense? 😅

resources/squadrons/super hornet/VFA-105F.yaml
Raffson, line 1: Any particular reason for completely removing this squadron? Wouldn't it make more sense to adjust the aircraft's name, if that's even necessary?

resources/squadrons/super hornet/VFA-41F.yaml
Raffson, line 1: Same here, any particular reason why the squadron got deleted? 😅

Raffson (changes requested):

pydcs_extensions/fa18efg/fa18efg.py
Drexyl: Have updated the weapons class in pydcs_extensions

resources/factions/blufor_current.json
Drexyl: Restored blufor_current (i think?)

resources/squadrons/super hornet/VFA-105F.yaml
Drexyl: Added squadrons back under their own folders which I think was my initial intention but left them out.

resources/squadrons/super hornet/VFA-41F.yaml
Drexyl: See above :)

pydcs_extensions/fa18efg/fa18efg.py
Raffson: Still contained some weapons that were declared more than once, but got rid of those locally so I'll push to your branch as soon as I finish up my review so we can merge ;)

resources/factions/blufor_current.json
Raffson: Not really, but I'm also having a tough time with this one, so I guess we'll just go with it 😅

Raffson (approved):

PR #491, opened 2025-03-20
EW Script 2.0 - Enable for AI

Drexyl:
Update for EW Script 2.0 implementation to allow certain AI flights to have the script enabled.

How it works...
* AI Defensive Jamming currently works for SEAD/SEAD Escort/SEAD Sweep.
* AI Defensive Jamming will start at JOIN Waypoint.
* If "ECM Required" option enabled, only AI with compatible ECM pod, as defined in /resources/weapons/pods will have Defensive Jamming Enabled. This also applies to Player aircraft.
* If "ECM Required" option is NOT enabled, all AI SEAD/SEAD Escort/SEAD Sweep flights will have Defensive Jamming enabled.

* For Player Aircraft, if the "ECM Required" option is NOT enabled, any player flight type will have Defensive/Offensive F10 Menu ability.

* Provisions have also been added for whitelisted aircraft to allow any aircraft to have Offensive Jamming enabled.
* Offensive Jamming for AI is currently only enabled for flights that use a Racetrack.


game/missiongenerator/aircraft/flightgroupconfigurator.py
```diff
@@ -173,21 +173,45 @@ def configure_flight_member(
...
+            return
+
+        # Check jammer requirement for non-specific aircraft types
+        if unit.type not in specific_aircraft_types:
+            if jammer_required:
+                ecm = WeaponType.JAMMER
+                if not member.loadout.has_weapon_of_type(ecm):
+                    return
+
+        if not member.is_player or not settings.plugins.get("ewrj"):
```
Raffson: Everything you seem to be setting up here is irrelevant since we'll always return in case of an AI, i.e. through the condition `not member.is_player`, but if I understood the rest properly, the setup for AI actually happens in the other files, so I think we can clean up most of this...

game/missiongenerator/aircraft/waypoints/joinpoint.py
```diff
@@ -9,16 +9,26 @@
...
+
 
 class JoinPointBuilder(PydcsWaypointBuilder):
     def add_tasks(self, waypoint: MovingPoint) -> None:
+        # List of specific aircraft types that should get their own ewrj_menu_trigger and be excluded from needing a jammer
+        specific_aircraft_types = ["CLP_E7A" , "CLP_P8" , "CLP_TU214R" , "CLP_TU214"]  # Replace with aircraft types e.g. E-3A
+
+        # List of excluded aircraft types that should not get any triggers
+        excluded_aircraft_types = ["F-16C_50"]  # Replace with aircraft types with working ECM
+
```
Raffson: Can we make these aircraft somehow configurable? The fact that we're repeating the same list across different files is not a good thing 😅
Also, noticing specific_aircraft_types` remains unused in joinpoint.py, which in a way makes sense 🤔

Raffson (commented):
Pushed a refactor, have a look at what happened and let me know if this works when you test it 😅

game/missiongenerator/aircraft/waypoints/joinpoint.py
Raffson: Another thing I'm starting to wonder about, why would we actually want to exclude aircraft-types? 😅

Drexyl: As discussed elsewhere, the excluded_aircraft_types was to account for modules that had working jammers. Maybe we just get rid of it?

> "Also, noticing specific_aircraft_types` remains unused in joinpoint.py, which in a way makes sense 🤔"
I only left this here for future use.

> "Can we make these aircraft somehow configurable?"
What's this "we" business? 😂
I did consider this, but it was a bit beyond my capabilities at this point in time. Definitely something I'd like to do in the future.

> "The fact that we're repeating the same list across different files is not a good thing 😅"
I do realize this, but again, beyond my current capabilities. 😂

Raffson (approved):

PR #504, opened 2025-04-27
Add Invisible FOBs

Druss99:
Adds invisible fob support, designated by a `Truck M939 Heavy`. Unit designation could change.

NOTE: Invisible FOBs won't have an ATC due to not having a Comms tower. This could be changed by using an SKP-11 or LUV HMMWV Jeep at the FOB but right now we have no logic in place that would allow for all factions to have access to one of those units purely for the purpose of FOB building. Logic could be added but I don't know if the ATC is particularly useful for a FOB and it also might break immersion for WWII campaigns to see a modern day HMMWV sitting there.

### DIFF
--- a/game/layout/layout.py
+++ b/game/layout/layout.py
@@ -288,6 +288,8 @@ class BuildingLayout(TgoLayout):
     @property
     def category(self) -> str:
         for task in self.tasks:
+            if task is GroupTask.INVISIBLE_FOB:
+                return "fob"
             if task not in [GroupTask.STRIKE_TARGET, GroupTask.OFFSHORE_STRIKE_TARGET]:
                 return task.description.lower()
         raise RuntimeError(f"Building Template {self.name} has no building category")
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # Retribution v1.5.0
 
 ## Features/Improvements
+* **[Campaigns]** Ability to define invisible FOBs
 
 ## Fixes
 
--- a/game/campaignloader/mizcampaignloader.py
+++ b/game/campaignloader/mizcampaignloader.py
@@ -53,6 +53,7 @@ class MizCampaignLoader:
 
     FOB_UNIT_TYPE = Unarmed.SKP_11.id
     FARP_HELIPADS_TYPE = ["Invisible FARP", "SINGLE_HELIPAD", "FARP"]
+    INVISIBLE_FOB_UNIT_TYPE = Unarmed.M_818.id
 
     OFFSHORE_STRIKE_TARGET_UNIT_TYPE = Fortification.Oil_platform.id
     SHIP_UNIT_TYPE = USS_Arleigh_Burke_IIa.id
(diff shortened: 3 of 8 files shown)


game/layout/layout.py
```diff
@@ -288,6 +288,8 @@ def create_ground_object(
     @property
     def category(self) -> str:
         for task in self.tasks:
+            if task is GroupTask.INVISIBLE_FOB:
+                return "fob"
```
Raffson, line 292: Shouldn't this be `invisiblefob`? or is `invisiblefob` a leftover? 😅

Raffson (commented):
Wouldn't we need something in the TgoGenerator to ensure in GroundObjectGenerator that the FOB is invisible? 🤔

game/layout/layout.py
Druss99: I have it returning fob here as the UI layer relies on task for the icons and there are probably other dependencies on the task. The invisible FOB should function identically as a normal FOB so reusing the task made sense to me

Druss99:
> Wouldn't we need something in the TgoGenerator to ensure in GroundObjectGenerator that the FOB is invisible? 🤔 

Unsure what you mean here unless you mean actually making the objects invisible. 

In the current implementation the FOB isn't actually invisible persay, it just only spawns a jerrycan (you have to spawn something) so that way you can put the FOB wherever you want without buildings colliding. If it's a FARP there will still be FARP statics for fuel and ammo but you need enough space for a helo already so that seems not problematic.

game/layout/layout.py
Raffson: Do we still need `invisiblefob` in building_data then? 😅

Druss99: I think we do because the group task determines what buildings get spawned and in order to only spawn a jerry can we need the `GroupTask.INVISIBLE_FOB`

Raffson:
> Unsure what you mean here unless you mean actually making the objects invisible. 

That's indeed what I mean, I think it makes sense to make the unit invisible, unless you see a problem with that approach? 😅

Also, are you sure that we absolutely _need_ to spawn a unit for a CP? Perhaps we can just use dummy data and skip the generation of the unit altogether?

Druss99:
> > Unsure what you mean here unless you mean actually making the objects invisible.
> 
> That's indeed what I mean, I think it makes sense to make the unit invisible, unless you see a problem with that approach? 😅
> 
> Also, are you sure that we absolutely _need_ to spawn a unit for a CP? Perhaps we can just use dummy data and skip the generation of the unit altogether?

So out of the FOB objects the only one that actually provides any functionality is the command post which provides you with an ATC. Everything else gets spawned in when you place helo pads (making it a FARP). I don't see any reason to make the objects invisible since they don't do anything (unless it's cleaner code wise). If we can make the command post invisible and still retain the functionality then we can just keep this same implementation, swap the jerrycan to a command post, and then make it invisible but IDK how invisible objects in DCS work, I thought they were inoperable when set to invisible.

Druss99:
If we can do some DCS lua magic to make all the FOB objects AND FARP objects invisible but they still work then that's definitely a better approach but also firmly outside of my knowledge of how DCS treats invisible objects. In fact I tested putting down all the objects and setting them to invisible and I was unable to rearm/refuel but maybe I was doing something wrong.

Raffson:
> If we can do some DCS lua magic to make all the FOB objects AND FARP objects invisible but they still work then that's definitely a better approach but also firmly outside of my knowledge of how DCS treats invisible objects. In fact I tested putting down all the objects and setting them to invisible and I was unable to rearm/refuel but maybe I was doing something wrong.

Ok I guess that rules it out then 🙈

Raffson (approved):

PR #518, opened 2025-05-28
Initial support for Germany Cold War

cedriclmenard:
Initial support for Germany Cold War

No campaign added with this PR, and PyDCS needs to be updated since airport frequency have changed in the latest patch (or so I've been told!). Will do ASAP.

The GIS steps were an absolute PITA with this one, given the amount of data and density in this region. Will try to do an update of the readme, since dissolving the polygons was just not possible at all. Had to instead make a raster-merge and then use some morphological operations on the raster image to achieve a somewhat small polygon file size.

The landmap file is still an order of magnitude bigger than the biggest one yet, so testing should be done to ensure this does not ruin performance, but this is outside of my current knowledge of the codebase haha!


resources/theaters/germanycw/info.yaml
```diff
@@ -0,0 +1,44 @@
...
+daytime:
+  dawn: [6, 9]
+  day: [9, 18]
+  dusk: [18, 20]
+  night: [0, 5]
+climate:
+  day_night_temperature_difference: 9.0
+  seasons:
+    winter:
+      average_pressure: 29.72 # TODO: Find real-world data
```
Raffson, line 13: This might be a good place to start: https://www.dwd.de/EN/ourservices/cdc/cdc_ueberblick-klimadaten_en.html

resources/theaters/germanycw/icon.gif
Raffson, line 1: The entire folder needs to be renamed to `germanycoldwar` instead

resources/tools/arcgis_landmap_import.py
```diff
@@ -13,16 +14,19 @@
...
-from game.theater import ConflictTheater, Landmap
-from game.theater.theaterloader import TheaterLoader
-from resources.tools.generate_landmap import to_multipoly
-
 THIS_DIR = Path(__file__).resolve()
 TOP_DIR = THIS_DIR.parents[2]
 IMPORT_DIR = TOP_DIR / "unshipped_data/arcgis_maps"
 RESOURCES_DIR = TOP_DIR / "resources"
 
+sys.path.append(str(TOP_DIR))
```
Raffson: Any particular reason you need this? Looks like an incomplete/incorrect setup 🤔

resources/theaters/germanycw/icon.gif
cedriclmenard: I think the actual name in the .miz/mission files is really "GermanyCW". Don't know if it makes a difference.

resources/tools/arcgis_landmap_import.py
cedriclmenard: Ah! Yeah, you're right. I think I didn't run the "Prepare Environment" task. Will remove and test!

cedriclmenard:
I've fixed the issues with naming. Map not showing the center on start is most likely due to an error on my part in pydcs, so opened a PR for that. Will need to update the commit hash for that once it's done. Thanks a lot for the review!

resources/theaters/germanycw/icon.gif
Raffson: You're right, it doesn't _really_ matter, it's rather of what we'd prefer to use in the campaign's yaml file, i.e. if we want to mention `GermanyColdWar` as Starfire did in his example, then we need to rename the folder, otherwise it's Starfire that needs to adjust his yaml to use `GermanyCW` instead 👍

cedriclmenard:
Just updated the landmap file with a clipped version of the exclusion zones based on the shared "Phase 3" screenshot of Germany Cold War in the newsletter. I think only the land and sea zones need to be updated with the same "borders" now, but at the same time they do cover all 3 phases even if they don't match the same borders as the exclusion zones.

And I think I've lost the sea and inclusion zones data 😅

Raffson:
To me it looks ok-ish, so let's roll the dice 🤞

Raffson (approved):

resources/theaters/germanycw/info.yaml
Raffson: Let's deal with this later...

PR #522, opened 2025-06-06
Add Moose as base plugin.

Drexyl:
This pull request is to add Moose as a base plugin.

This will allow the future use of Moose modules in Retribution.


Raffson:
I think it should still be seen as a separate lua-plugin because it's quite a large script, which would have an effect on the size of the generated missions

Raffson:
K well f*ck it, let's just merge this because otherwise this will just remain open for god knows how long 😅

Raffson (approved):

Issue #304, opened 2024-05-04
Waypoints in mission planner should start at 0 not 1.


41Mike304:
### Describe your proposal.

The generated flight plan in the mission planner creates a different waypoint structure to the generated kneeboard in game and lists "Takeoff" as WP1 followed by "hold" as WP2 etc, when it should be (as in the kneeboard)  "Takeoff" as WP0 and "Hold" as WP1 etc.

Raffson:
Can you provide some screenshots to indicate exactly what you're referring to? 😅

41Mike304:
Screenshots attached. This is the same flight plan shown in the flight planner and in game on kneeboard. 
![WP kneeboard](https://github.com/dcs-retribution/dcs-retribution/assets/137757411/1419f8a4-24a3-470d-a68e-b0becbec74a5)
![WP in editor](https://github.com/dcs-retribution/dcs-retribution/assets/137757411/7340b90a-cbd6-403a-8341-ce95e0483a14)

41Mike304:
This is the same flight plan shown in the flight planner and in game on kneeboard. As you can see, the generated in game kneeboard is correct listing takeoff as WP0. In the mission planner takeoff is listed as WP1.

![WP in KB](https://github.com/dcs-retribution/dcs-retribution/assets/137757411/19b08ffa-7737-4d00-a5a0-8b04a1630e5a)
![WP KB 2](https://github.com/dcs-retribution/dcs-retribution/assets/137757411/1a20b668-a8cb-4fd3-99b8-f2cb44ac8865)

41Mike304:
This issue has only been resolved in the edit WP box. It is still displaying the wrong WP structure on the map as seen in attached image. 

![Image](https://github.com/user-attachments/assets/bb99cc93-302a-4199-869b-74f6255a01b4)

PR #524, opened 2025-06-09
Fix issue 304: Waypoint numbers should start at 0 not 1 on map.

MrPozor:
Subtract 1 from the index to match the waypoint numbering of the waypoint list. Hope this is a good way of fixing this.
See screenshot showing that the numbering is now identical.
Screenshot 2025-06-09 123154


Raffson (approved):
Resolves #304

PR #525, opened 2025-06-10
Revert "Fix issue 304: Waypoint numbers should start at 0 not 1 on map."

MrPozor:
Reverts dcs-retribution/dcs-retribution#524

The fix creates a new problem when dragging the waypoint. We need another solution. Please roll back.


Raffson (approved):
Understood, I'll try to have z closer look at this one tomorrow 🤞

PR #540, opened 2025-07-14
Add support for Control Point Influence Zones

Druss99:
This PR adds support for defining control point influence zones. Control point influence zones are trigger zones defined in the campaign miz file that determine what units are owned by a CP. This allows for finer control over unit placement since you can now place a unit closer to an enemy CP than your own and still control it. Control point influence zones must be RED and must have the property value of the CP name.

*Dev note: I will take another look at this and potentially clean things up, I am pushing the PR now so @Starfire13 can test if he wants

### DIFF
--- a/game/campaignloader/mizcampaignloader.py
+++ b/game/campaignloader/mizcampaignloader.py
@@ -16,6 +16,7 @@ from dcs.terrain import Airport
 from dcs.unitgroup import PlaneGroup, ShipGroup, StaticGroup, VehicleGroup
 from dcs.vehicles import AirDefence, Armor, MissilesSS, Unarmed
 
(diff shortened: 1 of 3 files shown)


Druss99:
Already found a couple of changes that need to be made, namely restoring old behavior for carriers and not requiring influence zones for every CP. Will push tomorrow.

Druss99:
This should now be ready to test and merge. All my testing looks good. CPInfluenceZones are optional and do not need to be specified for every control point, you can specify them for as many or as few CPs as you would like.

Druss99:
Changed fallback behavior, now if a unit is not in an influence zone it will fall back to the closest CP without an influence zone. If all CPs have influence zones and a unit is outside of all of them it will raise a runtime error.

Raffson (approved):

PR #541, opened 2025-07-17
Essex as AircraftCarrier hack

Drexyl:
Updates Essex class carrier, class and tasks from HeliCarrier to AircraftCarrier.


Raffson:
But why? 😅
It's not as if you can spawn with a Hornet on the deck 😂

EDIT: or can you? 🤔

Raffson (approved):

PR #550, opened 2025-08-02
Add Moose + Soundhandler

StillClock1:
My first pull request contains two Plugins:

1. Moose
2. Soundhandler (which requires Moose) is intended to play sound effects based on weapon employment and different types of Kills (Air, Ground, Static).
Soundhandler uses EVENTS and plays sounds to the group that fires a weapon (i.e. Fox 3 sound for a Fox 3 shot, Pickle, Paveway, Magnum...etc.).  If the subject group is on the receiving end (i.e. Red air shoots missile at you or a SAM), the sound will only be played to your group, and with a randomized delay (3-7 seconds).

Includes its own F10 menu that allows On/Off toggle as well as an option to play all sounds to all of the same coalition, rather than just the involved group.


resources/plugins/MOOSE/plugin.json
Raffson, line 1: I think this can be left out since it's already part of the base, right? 😅

StillClock1: Yes, I didn't know what I needed to upload so I uploaded Moose so that the user had the prerequisite for Soundhandler.  Around that same time, Drexyl made a PR to include Moose as a base plugin.

My Moose file might be later than his, I don't know.

Feel free to comment out or remove extra lines or logging in Soundhandler.

StillClock1: What's the process, should I resubmit with just the Soundhandler plugin now?

PR #552, opened 2025-08-03
fix bugs with air wing config loading

Druss99:
Fixed 2 bugs with air wing config loading:
1. Carriers are now found via full name when loading which is how they are defined in the campaign yaml and how they are saved in air wing config files
2. When you load an air wing config all currently used squadrons are now unclaimed so if you are loading an air wing config that contains a squadron already present in the current air wing you can reuse it

### DIFF
--- a/game/theater/conflicttheater.py
+++ b/game/theater/conflicttheater.py
@@ -292,6 +292,12 @@ class ConflictTheater:
                 return cp
         raise KeyError(f"Cannot find ControlPoint named {name}")
 
+    def control_point_by_full_name(self, full_name: str) -> ControlPoint:
+        for cp in self.controlpoints:
+            if cp.full_name == full_name:
+                return cp
+        raise KeyError(f"Cannot find ControlPoint with full name {full_name}")
+
     def find_carriers(self) -> List[ControlPoint]:
         try:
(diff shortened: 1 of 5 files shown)


Druss99:
added aircraft_type to airwing config as a better fallback mechanism if the squadron is not found. aircraft_type is optional to not break save compat

game/theater/conflicttheater.py
```diff
@@ -292,6 +292,12 @@ def control_point_named(self, name: str) -> ControlPoint:
                 return cp
         raise KeyError(f"Cannot find ControlPoint named {name}")
 
+    def control_point_by_full_name(self, full_name: str) -> ControlPoint:
```
Raffson, line 295: Why do we need this method exactly? Afaict there's no difference with `control_point_named` because `full_name` and `name` ought to contain the same value, unless I'm missing something? 😅
I guess we should look for a different property as a fallback 🤔

Druss99: full_name contains the name set in the campaign yaml for carriers, name will result in the actual picked carrier name IE George-Washington, etc.

Raffson (approved):

PR #559, opened 2025-08-08
Autolase Plugin (Moose) - Looking for help flowing UI variables into plugin

StillClock1:
Options set in the Retribution UI seem to be overwritten by the variable set at the top of the file 'Plugin_Autolase_JTAC.lua'.

Can someone please take a look?

Requires Moose, not sure if anything is needed for that now.


StillClock1:
A feature of this Plugin is that the JTAC will orbit its current target.  If use the MarkerOps plugin you can destroy each smoked target with the F10 map markers and watch it adjust its orbit.  Helps prevent losing LOS.

resources/plugins/Autolase/plugin.json
```diff
@@ -0,0 +1,59 @@
...
+            {
+            "nameInUI": "JTAC Bravo - Notify Radius (NM)",
+            "mnemonic": "JtacBravoRadiusNM",
+            "defaultValue": 40
+    }
+    ],
+    "scriptsWorkOrders": [
+        {
+            "file": "Plugin_Autolase_JTAC.lua",
+            "mnemonic": "Autolase 0.1"
```
Raffson: I believe this is what's causing the issue... you can't really reference this in the config because it contains a space and a dot in the name, so I recommend changing this to `MooseAutolase` or something, then adjust the names on the right-hand side of the assignments, i.e. `dcsRetribution.plugins.MooseAutolase.JtacAlphaSmoke`

StillClock1:
Updated but same issue, settings changes made in the UI not flowing through based on the messages I see in the dcs.log of what those variable values are.

resources/plugins/Autolase/plugin.json
StillClock1: I updated this, no change - UI checkboxes not flowing into code.  The settings at the top of the actual file appear to override UI settings.  What am I missing?

StillClock1:
Branch superseded by new PR.

PR #560, opened 2025-08-08
MarkerOps Plugin

StillClock1:
This Plugin uses Moose to allow users in-game control using the F10 map markers and a few words of text.  There is no downside and just offers you optionality if your wing or mission calls for it.

1. "Boom" and a quantity (e.g. 'Boom 1000') into a map marker will trigger a 1,000 TNT-equivalent explosion at that coordinate.  Effect happens when F10 map marker is minimized.  No default explosion size, requires a number.  500 is enough to kill a tank, but a ship would require >2000.
2. "Light" drops 3 illumination flares over a coordinate at an altitude of 350 meters.
3. "Light Zone" or "Light zone" will create a 5 mile circle (by default) and drop illumination flares over every group, regardless of coaltion in that zone.  If you use the zone parameter it will also look for a number, so if you do "Light Zone 2" it will be a 2 mile circle.
4. "Smoke [color]" will place smoke of your chosen color at that coordinate (e.g. "Smoke red" or "Smoke "Red") will place red smoke on those coordinates for the default length of 5 min.  I can add sound effects to this if there's interest.
5. "SmokeZone" will smoke all groups in a 5 mile zone.  Blue units will smoke Blue, Red will smoke Red, and neutral will smoke white.  Similar to Light Zone above, this is a 5 mile zone by default but any number entered will become the radius.
6. "Destroy" will use Moose to destroy all groups (client and AI alike) in a 5 mile radius of your F10 map marker.  Enter a number after Destroy to widen or narrow the radius.
7. "RTB" will tell all AI groups in a 5 mile zone to immediately RTB.  Adjustable the same as above.
8. "Immortal" will make all groups in a 5 mile radius Immortal.  Adjustable the same as above.
9. "Mortal" will make all groups in a 5 mile radius Mortal.  Adjustable the same as above.
10. "Refuel" will tell all jets in a 5 mile radius to refuel at the nearest tanker, if they can.  Adjustable the same as above.


resources/plugins/MarkerOps/plugin.json
```diff
@@ -0,0 +1,13 @@
+{
+    "nameInUI": "MarkerOps (Requires MOOSE)",
+    "defaultValue": true,
+    "specificOptions": [],
+    "scriptsWorkOrders": [
+        {
+            "file": "Retribution_MarkerOps.lua",
+            "mnemonic": "MarkeOps 0.1"
```
Raffson: While this plugin doesn't have any specific options that are set through a config-file, it doesn't really break anything but I'd prefer to call thing `MooseMarkerOps` or something in that realm, basically only using alphanumeric characters and underscores 😅

StillClock1: Thanks, I've changed it to 'MooseMarkerOps' everywhere to make it easy.

Raffson (approved):

PR #562, opened 2025-08-10
expend all ARMs for SEAD

Druss99:
Expend all ARMs during SEAD enroute task to fix flights firing one ARM and RTB

Ref: https://wiki.hoggitworld.com/view/DCS_task_engageGroup

### DIFF
--- a/game/missiongenerator/aircraft/waypoints/seadingress.py
+++ b/game/missiongenerator/aircraft/waypoints/seadingress.py
@@ -62,6 +62,7 @@ class SeadIngressBuilder(PydcsWaypointBuilder):
                 # into the SAM instead of waiting for it to come alive
                 engage_task = EngageGroup(miz_group.id)
                 engage_task.params["weaponType"] = DcsWeaponType.ARM.value
+                engage_task.params["expend"] = Expend.All.value
                 waypoint.tasks.append(engage_task)
 
             # Use other Air-to-Surface Missiles at last
--- a/changelog.md
+++ b/changelog.md
@@ -18,6 +18,7 @@
 * **[Flight Plans]** Fixed a bug when a package was created with only escort flights
 * **[Flight Plans]** Added AntiShipStrike as a fallback task for OCA/Aircraft to fix a bug where the S-3B could not do OCA/Aircraft
 * **[Squadrons]** Fixed a bug where loading an air wing config would not properly load all squadrons
+* **[Flight Plans]** Fixed a bug where SEAD flights would fire one ARM and RTB
 
 # Retribution v1.4.1 (hotfix)
 


game/missiongenerator/aircraft/waypoints/seadingress.py
```diff
@@ -62,6 +62,7 @@ def add_tasks(self, waypoint: MovingPoint) -> None:
                 # into the SAM instead of waiting for it to come alive
                 engage_task = EngageGroup(miz_group.id)
                 engage_task.params["weaponType"] = DcsWeaponType.ARM.value
+                engage_task.params["expend"] = Expend.All.value
```
Raffson, line 65: This forces all HARMs to be fired only at 1 target, which I don't think is what we want, right? 😅

Druss99: Since this is "engageGroup" I think it should fire against all radars in the group, may need to test against a site with multiple radars like an SA-11

Raffson: And that's where this particular option will make the AI launch everything against a single unit, which is fine if group-attack is not turned on, although then you're still not sure that all radars will be covered 😅

Raffson: Actually... you might be right... 🤔
but still, I'd definitely double-check the behavior 😅

Druss99: Hmm I will test against an SA-11 but according to @Starfire13 with the current implementation they just fire 1 harm and leave regardless of if that harm is successful or if there are more radars. Group attack isn't enabled for this steer point so I still think it is net positive but will test further

Raffson: The better solution to that is to put the waypoint closer to the target, but then it's almost a certainty that the AI gets shot down, unless ED did something to improve their evasive maneuvers 🤔

Starfire13: They do fire at more than one radar against SA-11s, though won't fire at all of them. But I think that's still better than firing one HARM and RTBing.

Generally, I don't rely on HARM SEAD alone against SAMs with multiple radars. You either TALD SEAD and then DEAD them 3 minutes later, or just follow up the SEAD flight with a SEAD Sweep to clean up the remaining radars.

The main thing is to have SEAD working well against LORADS and the single radar MERADS (where saturation attacks are necessary). For pretty much everything else, SEAD Sweep works better.

Druss99: In my test against an SA-11 a 4 ship of F-16s with 4 HARMs each destroyed every single radar and against a patriot site protected by two SA-8s (all patriot launchers removed so they could actually hit the targets) they fired 3 harms initially and took out 3 separate radars and then later the rest of the flights fired all their harms against the remaining radars. 

If there are less radars than they have harms they seem to fire multiple at the same target but if there are multiple radars up they will fire at all of them not just fire all harms at one. 

This seems like a huge improvement in behavior and they do seem to actually try and saturate the radar with missiles.

Raffson (approved):

PR #565, opened 2025-08-22
New Moose-based Autolase alternative with UI options

StillClock1:
Moose-based alternative to CLTD autolase and vanilla-JTAC system.

Features:
1. Threat-sorting.
2. Moose's built-in autolase menu to change frequency, smoke color...etc
3. Default IR pointer.
4. JTAC Drone will circle current target (reducing target-lost situations) and move to next observed target.
5. Smoke deployed at offset to target to avoid obscuring IR weapon views.
6. Messages only to clients inside of configurable radius (default 40 miles) AND on one of two configurable UHF/VHF frequencies.  Clients receive both an audio cue and then a message containing laser frequency, (configurable) coordinates, and target type.
7. Supports Blue 'JTAC Alpha' and 'JTAC Bravo'
8. Built-In 'Danger Close' function, if there are friendlies within 1 (configurable) mile of the smoked target, a different sound is triggered and the friendly will deploy Green or Blue smoke (random) to ensure no friendly fire.

Planned Updates:
1. More configurable options.
2. Support beyond just 'JTAC Alpha' and 'JTAC Bravo'


resources/plugins/MooseAutolase/plugin.json
```diff
@@ -0,0 +1,60 @@
+{
+    "nameInUI": "Moose Autolase",
+    "defaultValue": true,
```
Raffson: Wouldn't this conflict with CTLD's autolase? 🤔

StillClock1: Yes, it is intended as an alternative and would very likely conflict if both run at the same time.  I don't think its specific to those three lines your post references though.

For me, the constant flashing of messages was the main reason I wanted to give myself an alternative.  Also my wing observed issues with CLTD's autolase not moving to the next target - wasn't me personally so I can't confirm.

Raffson: But then it would make sense to set CTLD's autolase default value to false, or keep this one false, and override this with default settings for you locally?

Raffson: Fyi, branch also has merge-conflicts since one of your PRs got merged 😅

StillClock1: Should be all set now, I've rebased to the latest dev.

StillClock1: > But then it would make sense to set CTLD's autolase default value to false, or keep this one false, and override this with default settings for you locally?

Yes - default is now false.

Raffson (approved):

PR #567, opened 2025-08-29
fix naval squadron retreat bug

Druss99:
Fix a bug where destroying a carrier group escort would cause all squadrons on the carrier to disband


Raffson (approved):
Nice catch 👍

Issue #561, opened 2025-08-09
Frontline unit naming request


StillClock1:
### Is your feature request related to a problem? Please describe.

I would like to try the Tiresias module (4 lines of code) that turns off the AI for ground units far from any opposing aircraft or helo (e.g. the groups positioned around control points but not on the frontlines).  Tiresias does not affect ships, SAMs, or planes and could save large amounts of CPU and memory.

### Describe the solution you'd like

Append a consistent string to all frontline GROUP names, I recommend "Frontline".  All convoy units already have 'Convoy' as part of their group name so its easy to add an exception set for them.  The code looks like this:

'''lua
local exceptionset = SET_GROUP:New():FilterPrefixes({ "Convoy", "convoy" }):FilterStart()
local blinder = TIRESIAS:New()
--  Setup different radius for activation around helo and airplane groups (applies to AI and humans)
blinder:SetActivationRanges(10, 25) --  defaults are 10, and 25
--  Setup engagement ranges for AAA (non-advanced SAM units like Flaks etc) and if you want them to be AIOff
blinder:SetAAARanges(60, true)      --  defaults are 60, and true
blinder:AddExceptionSet(exceptionset)
'''

StillClock1:
Another idea would be "TIC" so the Troops in Contact script could be made a plugin.  I don't need any specific string, it just needs to be consistent across the frontline units.

cedriclmenard:
I've done a draft PR with the proposed changes. I think there is a better way to implement it, but in the meantime if you want to test, it is available. I won't be available for the next 2 weeks or so, but will gladly take a look back and maybe implement a better way of doing this or some possible fixes when I'll be back! :)

StillClock1:
Thanks for doing this, I'll play with it and let you know what I see.  My test method last time was: 
1. Are convoys moving?  Under my previous code - Yes.
2. Are frontline units moving? Under my previous code - Yes.
3. Are frontline units engaging each other? Under my previous code - No.
4. Are SAMs firing when they should? - not yet tested, but SAMs are excluded by default for Tiresias (just like jets and boats)

PR #568, opened 2025-08-30
Added implementation of Tiresias according to StillClock1 (Issue #561)

cedriclmenard:
Basic implementation of the idea by @StillClock1 (Issue #561).

I think the best way to do this would be to instead feed the unit names directly from Retribution to the LUA script via the plugin interface (a bit like how it is done for the artymbot plugin), but this should suffice for some testing (hence the draft PR).

I tested it, but there seems to be some issues with unresponsive AIs on the frontline nonetheless. I will be doing some more investigation, but in the meantime, feel free to test the performance difference.

### DIFF
--- a/game/missiongenerator/flotgenerator.py
+++ b/game/missiongenerator/flotgenerator.py
@@ -158,7 +158,7 @@ class FlotGenerator:
             country = self.mission.country(self.game.blue.faction.country.name)
             jtac = self.mission.flight_group(
                 country=country,
-                name=namegen.next_jtac_name(),
+                name=namegen.next_jtac_name() + " FRONTLINE",
                 aircraft_type=utype.dcs_unit_type,
                 position=position[0],
                 airport=None,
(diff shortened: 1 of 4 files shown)


Drexyl:
Try this...

```env.info("-----DCSRetribution|MOOSE Tiresias plugin - configuration start ------")

local exceptionset = SET_GROUP:New():FilterPrefixes({ "unit|80", "unit|81", "JTAC", "Convoy" }):FilterStart()

-- local exceptionset = SET_GROUP:New():FilterPrefixes({ "JTAC" }):FilterStart()

for _, group in pairs(exceptionset:GetSet()) do
    env.info("TIRESIAS Exception Group: " .. group:GetName())
end

local blinder = TIRESIAS:New()
-- Setup different radius for activation around helo and airplane groups (applies to AI and humans)
blinder:SetActivationRanges(100, 150) -- defaults are 10, and 25
-- Setup engagement ranges for AAA (non-advanced SAM units like Flaks etc) and if you want them to be AIOff
blinder:SetAAARanges(60, true) -- defaults are 60, and true
blinder:AddExceptionSet(exceptionset)```

PR #575, opened 2025-09-19
Updated soundhandler with mirror performance for Red and Blue Clients…

StillClock1:
Update to Soundhandler:
1. Mirrored sounds for Red and Blue, suitable for PvP play.
2. Removed sounds played when friendly unit killed of same coalition (i.e. on frontline) as there are too many.
3. Removed OpFor gun shooting sound, need reliable way to determine its target - TBD.
4. Greatly removed logging if Debug is off.


resources/plugins/MooseSoundhandler/plugin.json
```diff
@@ -1,10 +1,10 @@
 {
-    "nameInUI": "Soundhandler",
-    "defaultValue": false,
+    "nameInUI": "Moose Soundhandler",
+    "defaultValue": true,
```
Druss99: IMO we should keep this to false, almost all Lua scripts bundled with retribution are set to false by default since they are there to customize your experience.

StillClock1: Agreed, I'll change that.

StillClock1: PR updated

Druss99:
Need to check if this will break save compat, might get to it tomorrow but we will see

StillClock1:
> Need to check if this will break save compat, might get to it tomorrow but we will see

Thanks.  Take your time, only if its something you'd enjoy having.  Have you done a flight with it on before?

Druss99 (approved):

PR #593, opened 2025-10-18
Update SplashDamage script to 3.4.2

Drexyl:
Update Splash Damage Script to 3.4.2 - https://forum.dcs.world/topic/370261-splash-damage-3x-script-now-with-more-explosions-version-34-now-with-more-stuff/

Added Ghosti/Raffson additions of OCA Multiplier and Ship Radar Damage from previous versions.

### DIFF
--- a/game/game.py
+++ b/game/game.py
@@ -4,6 +4,7 @@ import itertools
 import logging
 import math
 from collections.abc import Iterator
+from copy import deepcopy
 from datetime import date, datetime, time, timedelta
 from enum import Enum
 from typing import Any, List, TYPE_CHECKING, Union, cast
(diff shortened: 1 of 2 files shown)


Druss99:
Can we instead support both versions? Some people might prefer the new but others might prefer the old version

Starfire13:
I think if this already includes all the additions Stoner and Ghosti made to SP2, then it's not so bad if it's just a straight up replacement. I see that this version has optional cluster bomb support that is disabled by default. Is that toggleable via plugin options? Are the two new features (cook off and ordnance protection) similarly toggleable? If so, then I think we can just use SP3 and the user can pick and choose what options he enables.

Drexyl:
> Can we instead support both versions? Some people might prefer the new but others might prefer the old version

As discussed on Discord, SD3 is actually a fork of SD2, so no need to support both.



> I think if this already includes all the additions Stoner and Ghosti made to SP2, then it's not so bad if it's just a straight up replacement. I see that this version has optional cluster bomb support that is disabled by default. Is that toggleable via plugin options? Are the two new features (cook off and ordnance protection) similarly toggleable? If so, then I think we can just use SP3 and the user can pick and choose what options he enables.

I added the Ship radar damage from HARM, and the OCA aircraft multiplier from previous PR's for SD2. However cluster munitions are now handled by SD3, and not Ghosti's SD2 implementation.

Cluster munitions were disabled by default for SD2, I have kept this same behavior for SD3 with a toggle switch to enable it.

Options for new features like cook off and ordinance protection to be implemented in the future.

PR #607, opened 2025-11-04
10.8 Update for Noisy Cricket, Aegean Aegis, and Vegas Nerve

Starfire13:
I also updated the Iran 2015 faction as it was missing a couple of choppers. Added in the Hip, SeaCobra, and Huey.


resources/campaigns/exercise_vegas_nerve.yaml
```diff
@@ -91,7 +96,7 @@ squadrons:
       secondary: any
       aircraft:
         - C-130J-30 Super Hercules
-        - CH-47F Block I
+        - C-130
```
Druss99, line 99: I see the C-130 is already listed above was this a typo?

Druss99 (changes requested):

resources/campaigns/exercise_vegas_nerve.yaml
Starfire13: No. This picks the c-130J mod aircraft as the primary choice but falls back to the non-mod version for those who aren't using the mod.

Druss99 (approved):

PR #608, opened 2025-11-05
Swap F/A-18 intra-flight and inter-flight radio indices to accurately reflect real-world procedures

Starfire13:


Druss99:
Whoever did this originally left a comment that radio 1 needs to be used for intra flight so I might need to look at this more and see what DCS allows for the two radios

Druss99 (approved):

PR #612, opened 2025-11-07
Fixed bugs wrt planning escort flights

Druss99:
Fixes infinite recursion mentioned here: https://discord.com/channels/1015931619187621999/1025835702111445112/1435991888439279727

Also fixes deleting flights not recreating the package, this should happen since deleting a flight can change the primary flight.

Also fixes package validation not properly deleting the flights when the package gets marked as invalid.

### DIFF
--- a/qt_ui/windows/mission/QPackageDialog.py
+++ b/qt_ui/windows/mission/QPackageDialog.py
@@ -298,6 +298,8 @@ class QPackageDialog(QDialog):
         ):
             return True
         else:
+            for flight in list(self.package_model.package.flights):
+                self.package_model.cancel_or_abort_flight(flight)
             QMessageBox.critical(
                 self,
                 "Invalid Package",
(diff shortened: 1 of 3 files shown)


qt_ui/windows/mission/QPackageDialog.py
```diff
@@ -239,6 +239,16 @@ def on_delete_flight(self) -> None:
...
+            flight.recreate_flight_plan()
+            self.package_model.update_tot()
+            EventStream.put_nowait(GameUpdateEvents().new_flight(flight))
+        except PlanningError as ex:
+            self.package_model.add_flight(flight)
+            logging.exception("Could not delete flight")
+            QMessageBox.critical(
+                self, "Could not delete flight", str(ex), QMessageBox.StandardButton.Ok
+            )
         self.package_model.cancel_or_abort_flight(flight)
```
Raffson, line 242: Shouldn't this happen before the try-except? 🤔

Druss99: I copied this directly from the code above a few errors were made I think

PR #619, opened 2025-11-23
Updating the HighDigitSAMs mod extension to 2.1.0.

Biggus22:
As per title, I've added missing units from the 2.1.0 release of the High Digit SAMs mod.


Raffson (commented):
Could you elaborate what these new options do for the Phantom? 😅

Other than that this PR is what I would expect 👍

Biggus22:
> Could you elaborate what these new options do for the Phantom? 😅
> 
> Other than that this PR is what I would expect 👍

Ah damn, thought I had included it in my first comment.  I seem to have bundled them together in this PR.

Combat Tree is or was a system that would fool Soviet transponders into sending a response, giving away their position. Heatblur added it a few patches ago. It has three options: off, historical (which limits it to aircraft it actually worked on) and all.
The EFB is the electronic flight bag, which is basically a neat pop up providing some cool data about aircraft systems.

Biggus22:
I think I've messed up my repo a bit and I suspect this is a remnant of my working on a different project within Retribution where I've pulled or pushed something I shouldn't have.

PR #634, opened 2025-12-28
JTAC aircraft painter

MetalStormGhost:
Implemented AircraftPainterJtac which applies faction liveries_overrides to JTAC units generated for eact FLOT.


game/missiongenerator/flotgenerator.py
```diff
@@ -165,6 +166,14 @@ def generate(self) -> None:
...
             )
+            AircraftPainterJtac(self.game.blue.faction, utype, jtac).apply_livery()
+            jtac.points[0].tasks.append(
+                FAC(
+                    callsign=len(self.mission_data.jtacs) + 1,
+                    frequency=int(freq.mhz),
+                    modulation=freq.modulation,
+                )
+            )
             cs = jtac.units[0].callsign_dict
```
Raffson, line 170: This will result into 2 FAC tasks, so I think this can be removed cause at line 180 we're doing the same thing 🤔

game/missiongenerator/flotgenerator.py
```diff
@@ -165,6 +166,14 @@ def generate(self) -> None:
...
                 maintask=AFAC,
             )
+            AircraftPainterJtac(self.game.blue.faction, utype, jtac).apply_livery()
+            jtac.points[0].tasks.append(
+                FAC(
+                    callsign=len(self.mission_data.jtacs) + 1,
+                    frequency=int(freq.mhz),
+                    modulation=freq.modulation,
+                )
+            )
```
Raffson: This will result into 2 FAC tasks, so I think this can be removed cause at line 180 we're doing the same thing 🤔

MetalStormGhost: I have now removed the duplicate FAC task.

Raffson (approved):

PR #637, opened 2025-12-31
Update SplashDamage3 defaults and add new options

Drexyl:
This PR brings SpashDamage3 defaults back to more realistic levels without reducing functionality.
Added additional options for ordnance protection and blast wave from large explosion multipliers among others.


resources/plugins/splashdamage3/plugin.json
```diff
@@ -52,16 +52,16 @@
...
         },
         {
-            "nameInUI": "Overall Scaling (%)",
+            "nameInUI": "Overall Scaling Multiplier",
             "mnemonic": "overall_scaling",
             "minimumValue": 1,
-            "maximumValue": 5000,
-            "defaultValue": 100
+            "maximumValue": 100,
+            "defaultValue": 1
```
Druss99: My only concern here is that we have this setting which acts as a multiplyer but in line 25 of the config.lua we divide by 100:
```
splash_damage_options.overall_scaling = dcsRetribution.plugins.splashdamage3.overall_scaling / 100
```

My assumption is that it wasn't properly being divided by 100 or that something else was going on.

I think before merging this maybe we should add some logging lines to the script to see what value is actually getting used in the splash damage plugin once it starts. We might be applying a multiplier of .001 in which case were essentially running with no splash damage script at all.

Raffson: Since we're reading this as an integer, making this change will mean that you can't scale down below 1. That's why the overall scaling is considered to be a percentage, since that gives you the ability to have a multiplier as low as 0.01. I don't think we'll need more granularity for these options, hence the division by 100.
Does that make sense? 😅

PR #641, opened 2026-01-10
Update CH Russia pack to 2.0

Druss99:


game/dcs/groundunittype.py
```diff
@@ -109,6 +109,18 @@ def each_dcs_type() -> Iterator[Type[VehicleType]]:
     def _data_directory(cls) -> Path:
         return Path("resources/units/ground_units")
 
+    @staticmethod
+    def _migrator() -> Dict[str, str]:
+        return {
+            "[CH] T-90A MBT": "MBT T-90M [CH]",
+            "[CH] T-90M MBT": "MBT T-90M [CH]",
```
Druss99, line 116: I want to note that I am making this change here as opposed to adding the variant name to the existing unit yaml because I want it to be more clear which units are a part of the CH mod pack vs CH native units. CH native units have the [CH] at the end while mod units have the [CH] at the front.

Raffson (approved):

PR #642, opened 2026-01-11
prevent ground objects from being owned by off map spawns

Druss99:
This fixes a bug where an off map spawn could have control over a TGO object which would cause the TGO to not spawn at all in the campaign

### DIFF
--- a/game/campaignloader/mizcampaignloader.py
+++ b/game/campaignloader/mizcampaignloader.py
@@ -546,7 +546,12 @@ class MizCampaignLoader:
                 not in [
                     ControlPointType.AIRCRAFT_CARRIER_GROUP,
                     ControlPointType.LHA_GROUP,
+                    ControlPointType.OFF_MAP,
                 ]
+            ]
+        else:
+            candidates = [
+                cp for cp in candidates if cp.cptype != ControlPointType.OFF_MAP
             ]
 
         if candidates:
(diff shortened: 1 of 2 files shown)


game/campaignloader/mizcampaignloader.py
```diff
@@ -546,8 +546,13 @@ def objective_info(
                 not in [
                     ControlPointType.AIRCRAFT_CARRIER_GROUP,
                     ControlPointType.LHA_GROUP,
+                    ControlPointType.OFF_MAP,
```
Raffson, line 549: Perhaps the better way to handle this is to adjust the influence zone? 🤔

Raffson: Suppose we filter out the influence zones that are tied to an off-map spawn, wouldn't that be cleaner? 😅 

```
    @cached_property
    def cp_influence_zones(self) -> List[ControlPointInfluenceRadius]:
        return [
            zone
            for zone in ControlPointInfluenceRadius.from_trigger_zones(
                self.mission.triggers._zones
            )
            if self.theater.control_point_named(zone.cp_name).cptype
            not in [ControlPointType.OFF_MAP]
        ]
```
Unless an off-map spawn has a need for an influence zone, I think it's probably better to eradicate them altogether from the influence zones. Do you by any chance know of a case where the influence zone would be needed for off-map spawns? 😅

Druss99: I would argue that this way is better since it keeps all of the off map spawn filtering in the same place as opposed to being abstracted away in a different method. If you read the code from line 549 to 581 it is very clear that off map spawns are not allowed to own TGOs but if I moved it you might think on first glance that they would be allowed if they had an influence zone. Up to you but I think it is more readable living here even if it is technically more verbose

Raffson: Agreed, at first I thought it was possible to tuck everything away nicely, but that's not even the case when I take a closer look 😅

Raffson (approved):

PR #643, opened 2026-01-11
add campaign name to retribution window name

Druss99:
Add the name of the campaign to the name of the retribution window to make it easier for users to know what campaign they are using and make debugging easier.

### DIFF
--- a/game/game.py
+++ b/game/game.py
@@ -7,7 +7,7 @@ from collections.abc import Iterator
 from copy import deepcopy
 from datetime import date, datetime, time, timedelta
 from enum import Enum
-from typing import Any, List, TYPE_CHECKING, Union, cast
+from typing import Any, List, Optional, TYPE_CHECKING, Union, cast
 from uuid import UUID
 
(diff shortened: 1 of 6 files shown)


game/game.py
```diff
@@ -171,6 +173,8 @@ def __setstate__(self, state: dict[str, Any]) -> None:
             self.laser_code_registry = LaserCodeRegistry()
             for front_line in self.theater.conflicts():
                 front_line.laser_code = self.laser_code_registry.alloc_laser_code()
+        if not hasattr(self, "campaign_name"):
+            self.campaign_name = None
```
Raffson: This looks like something that should go to `migrator.py` 🤔

Raffson (commented):
Pretty much identical to what I had in mind, aside from the backwards-compatibility hack for `campaign-name` :wink:

game/game.py
Druss99: I think we don't even need a migration since it is an optional attribute but I am checking now

Druss99: updated

Raffson (approved):

PR #644, opened 2026-01-11
Vietnam War Vessels 2.3.0 Support

Biggus22:
This is a PR to update Vietnam War Vessels from 0.9.0 to 2.3.0.  Several new units have been implemented, some units have had name changes (in particular CVA-31 has been replaced "[VWV] CVA-31 Bon Homme Richard").  A few units have names that aren't entirely consistent with the previous implementation (A-37 and H-2 variants, probably others I can't recall right now).

Factions have been updated to reflect the changes.


Biggus22:
Deleting because I need to create a new branch.

PR #645, opened 2026-01-11
Updating Vietnam War Vessels to 2.3.0

Biggus22:
This adds Vietnam War Vessels 2.3.0 support, updated from 0.9.0.  Several units have had some name changes (CVA-31 in particular) and some units do not follow naming conventions in a few places.  Naming of many units is a bit inconsistent, so I've relied upon variant names in order to make things a bit more uniform where I could.


game/factions/faction.py
```diff
@@ -672,24 +672,50 @@ def apply_mod_settings(self, mod_settings: Optional[ModSettings] = None) -> None
             self.remove_vehicle("TR_TT")
             self.remove_vehicle("Gozanti")
             self.remove_ship("Destroyer_carrier")
-        # vietnamwarvessels
+        # vietnamwarvessels (using variant names because it was the closest thing to a consistent formula I could manage with this mod)
```
Raffson: This seems a bit strange cause `remove_ship` & `remove_aircraft` expect the DCS-ID, so you'd have to reference the id's in `vietnamewarvessels.py` 🤔

Biggus22: Hopefully I've sorted it out now! :)

resources/units/ships/USS Enterprise 1966.yaml
Druss99, line 1: Some unit files like this one include underscores but the DCS unit ID does not, they need to match the dcs unit ID to be picked up

Biggus22: Hopefully this resolves it.  Also I missed it on the first pass but I can't see the HH-2D appears to have been removed.

.gitignore
```diff
@@ -25,3 +25,5 @@ env/
 /resources/plugins/pretense/pretense_output.lua
 
 *.psd
+temp_run.py
+resources/factions/USA Vietnam Test.json
```
Raffson: You should probably remove this line & add your faction under `Saved Games > DCS > Retribution > Factions` 😅

Biggus22: I pulled the test faction out, it didn't really serve a suitable purpose outside of testing.

Raffson (approved):

PR #646, opened 2026-01-11
use batch writing for state.json

Druss99:
Use a batch writing strategy for the state.json file to reduce CPU utilization. Currently set to every 15 seconds if the state has changed and on mission end. This seemed like a good compromise between not losing much data if the server/game crashes but also not overburdening the CPU.

The biggest change is here:
```
        if destruction.type ~= nil and 
           string.find(destruction.type, "GENERIC_CRASH_MODEL") == nil and
           string.find(destruction.type, "_CRASH") == nil then
            destroyed_objects_positions[#destroyed_objects_positions + 1] = destruction
```

There was some concern that this might remove objects we actually care about but this gets saved here:
```
        return cls(
            mission_ended=data["mission_ended"],
            killed_aircraft=killed_aircraft,
            killed_ground_units=killed_ground_units,
            destroyed_statics=data["destroyed_objects_positions"],
            base_capture_events=data["base_capture_events"],
        )
```

And is used here:
```
    def record_carcasses(self, debriefing: Debriefing) -> None:
        for destroyed_unit in debriefing.state_data.destroyed_statics:
            self.game.add_destroyed_units(destroyed_unit)
```

For destroyed units we only care about units that resolve to a DCS unit type which these crash objects will not resolve to:
```
    def generate_destroyed_units(self) -> None:
        """Add destroyed units to the Mission"""
        if not self.game.settings.perf_destroyed_units:
            return

        for d in self.game.get_destroyed_units():
            try:
                type_name = d["type"]
                if not isinstance(type_name, str):
                    raise TypeError(
                        "Expected the type of the destroyed static to be a string"
                    )
                utype = unit_type_from_name(type_name)
            except KeyError:
                logging.warning(f"Destroyed unit has no type: {d}")
                continue
```

So this change should be safe and dramatically reduces lag and time to process the state.json.

### DIFF
--- a/resources/plugins/base/dcs_retribution.lua
+++ b/resources/plugins/base/dcs_retribution.lua
@@ -1,5 +1,5 @@
--- the state.json file will be updated according to this schedule, and also on each destruction or capture event
-local WRITESTATE_SCHEDULE_IN_SECONDS = 60
+-- the state.json file will be updated according to this schedule and on mission end
+local WRITESTATE_SCHEDULE_IN_SECONDS = 15
 
 logger = mist.Logger:new("DCSRetribution", "info")
 logger:info("Check that json.lua is loaded : json = "..tostring(json))
@@ -11,6 +11,7 @@ kill_events = {} -- killed units will be added via S_EVENT_KILL
 base_capture_events = {}
 destroyed_objects_positions = {} -- will be added via S_EVENT_DEAD event
 mission_ended = false
+dirty_state = false -- Track if state has changed and needs writing
 
 local function ends_with(str, ending)
    return ending == "" or str:sub(-#ending) == ending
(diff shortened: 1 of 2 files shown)


resources/plugins/base/dcs_retribution.lua
```diff
@@ -170,20 +175,23 @@ local function onEvent(event)
...
         write_state()
+        mission_ended = true
     end
 
 end
 
 mist.addEventHandler(onEvent)
 
--- create the state.json file and start the scheduling
+dirty_state = true
```
Druss99, line 196: Added so that we attempt to write the state at the start of the mission to catch any errors with being unable to write to the state file

Raffson (approved):

PR #652, opened 2026-01-18
improve logging for layout loading

Druss99:
Honestly I'm not sure if we should log an error or raise an exception for these issues but I am happy to adjust this PR as desired

### DIFF
--- a/game/armedforces/forcegroup.py
+++ b/game/armedforces/forcegroup.py
@@ -370,7 +370,19 @@ class ForceGroup:
                 else:
                     statics.append(static_type)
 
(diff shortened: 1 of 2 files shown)


Raffson:
> Honestly I'm not sure if we should log an error or raise an exception for these issues but I am happy to adjust this PR as desired

Raising an exception (which is currently the case) pretty much breaks all layouts, so I think what you cooked up here makes sense, though we _may_ want to mention that "import layouts" could potentially fix the problem :thinking:
My guess is that this error will most likely pop up when users are adding or adjusting custom layouts that they've made themselves, in which case a layouts import is usually required...

Issue #479, opened 2025-02-03
Escort flight improvements


Druss99:
### Is your feature request related to a problem? Please describe.

Right now there are 3 main issues:
- The join waypoint for escort flights is way to close to the ingress point and should instead be near the hold point. This is a little complicated to fix because currently the hold point is generated after the join waypoint
- Escorts disregard all threat so if you plan a refueling flight and then add escorts, the refueling flight will avoid enemy SAMs etc. but the escort flights waypoints will take them directly to wherever the target point is that the refueling flight was created on.
- Escort flights can be planned with no other flights in the package which results in a bug mentioned here: https://github.com/dcs-retribution/dcs-retribution/pull/477

### Describe the solution you'd like

Longterm we probably need to not allow manual planning of escorts and instead when you create a flight there should be a checkmark for enabling escorts, if you check it then it will bring up options for escort aircraft. Then it should generate the main flight and the escort flight gets generated afterwards, that way it can use the main flights waypoints as a basis for its own waypoints.

PR #657, opened 2026-01-23
Enhance join point selection with randomized distance

red-one1:
This PR moves the join/split points much closer to home - 25-40% of the distance to target.

This makes sure that escort flights are nicely formed up with their escortees, and haven't gotten shot down before forming up.

This should also solve (mostly?) #479


Raffson:
I don't think the location of the join-point should change, because generally speaking it should be closest to the "main" flight's departure, but the escorts need a better flight-plan between their departure and the join-point. Instead I think the escorts should plan a better route between their departure and the join-point 🤔

red-one1:
Currently the join point is pretty close to ingress (I'd say about 80% of the way from base to target) - if they're flying over a large territory, odds are they won't have escorts until they're deep into hostile airspace.
Hence this change - make sure they're escorted ASAP.

Raffson:
> Currently the join point is pretty close to ingress (I'd say about 80% of the way from base to target) - if they're flying over a large territory, odds are they won't have escorts until they're deep into hostile airspace.
> 
> Hence this change - make sure they're escorted ASAP.

Right, that rings a bell indeed 😅
I'll try to review this properly over the weekend 🤞

Starfire13:
Yes. This is something I've requested for a while now. Join waypoints right now are waaaaay too far from hold and too close to ingress, so you have to manually move them for each package, which is a pain. I wonder though, why the need for randomisation, rather than just a fixed %distance past Join?

red-one1:
> Yes. This is something I've requested for a while now. Join waypoints right now are waaaaay too far from hold and too close to ingress, so you have to manually move them for each package, which is a pain. I wonder though, why the need for randomisation, rather than just a fixed %distance past Join?

Just to make it interesting, TBH.

Raffson:
Ran a quick test to compare the current implementation against yours:
image

I'm getting a bit worried that some of the flight-plans are getting too long this way 😅
It would make more sense if the join point was somewhere near the circle I dropped in the screenshot.
The obvious question at this point is how do we fix this inefficiency 🤔

game/flightplan/joinzonegeometry.py
```diff
@@ -97,8 +114,30 @@ def __init__(
         self.preferred_lines = preferred_lines
 
     def find_best_join_point(self) -> Point:
-        if self.preferred_lines.is_empty:
-            join, _ = shapely.ops.nearest_points(self.permissible_zones, self.ip)
+        if random_point := self._random_point_in_geometry(self.permissible_zones):
+            return random_point
```
Raffson: I believe getting rid of this part fixes the issue I pointed out. Granted, the flight-plans get a _little_ more boring that way but the desired effect is established when I leave this part out:
image
image
image

Other than that I think you nailed it 👍

game/flightplan/joinzonegeometry.py
```diff
@@ -97,8 +114,30 @@ def __init__(
...
+            search_geometry = self.preferred_lines
+        elif not self.permissible_zones.is_empty:
+            search_geometry = self.permissible_zones
         else:
-            join, _ = shapely.ops.nearest_points(self.preferred_lines, self.ip)
+            # No usable geometry; fall back deterministically to the IP position.
+            join = self.ip
+            return self._target.new_in_same_map(join.x, join.y)
+
+        join, _ = shapely.ops.nearest_points(search_geometry, self.ip)
```
Raffson, line 127: If you want to add some randomization, you could introduce a small perturbation on the join-point. I'd probably go for a 10-15NM bubble around the join point while only considering the half which is closest to home-plate, which might be a little tricky, so a 5NM bubble will probably be fine too 😅 
That said, at this point I don't really think it's necessary either. When I compare the location of the join-points with the current behavior in `dev`, I think @Starfire13 is going to like this a lot 😁

game/flightplan/joinzonegeometry.py
```diff
@@ -97,8 +114,30 @@ def __init__(
...
         return self._target.new_in_same_map(join.x, join.y)
+
+    def _random_point_in_geometry(self, geometry: MultiPolygon) -> Point | None:
```
Raffson: If lines 117-119 get removed, then we can clean this up as well 😉

Raffson (changes requested):

red-one1:
> Ran a quick test to compare the current implementation against yours: image
> 
> I'm getting a bit worried that some of the flight-plans are getting too long this way 😅 It would make more sense if the join point was somewhere near the circle I dropped in the screenshot. The obvious question at this point is how do we fix this inefficiency 🤔


I don't understand - is your screenshot including my change, or not?

game/flightplan/joinzonegeometry.py
red-one1: Randomness removed.

game/flightplan/joinzonegeometry.py
red-one1: Done

game/flightplan/joinzonegeometry.py
red-one1: No randomness anymore - join/split point creation is now "the closest point (to INGRESS) in a band between 35-36% of the distance to target"

red-one1:
General comment - if escorts and bombers are taking off from different airfields, the bombers get to pick the join point, and the escorts will just have to deal with it.

Raffson:
> I don't understand - is your screenshot including my change, or not?

The first screenshot included all of your changes, the other ones were made after removing the "random point in geometry" method 😉

Raffson:
> General comment - if escorts and bombers are taking off from different airfields, the bombers get to pick the join point, and the escorts will just have to deal with it.

Correct, but escorts should be chosen in such a way that the closest available aircraft get picked. Not a lot we can do about it aside from shifting squadrons perhaps 😅

Raffson (approved):

PR #658, opened 2026-01-26
Add VSN F-35A/B/C mod support

Drexyl:
Adds VSN's F-35A/B/C mod support.


resources/units/aircraft/VSN_F35A.yaml
```diff
@@ -0,0 +1,49 @@
...
+  intra_flight: AN/ARC-210
+  inter_flight: AN/ARC-210
+  channels:
+    type: common
+    # DCS will clobber channel 1 of the first radio compatible with the flight's
+    # assigned frequency. Since the F/A-18's two radios are both AN/ARC-210s,
+    # radio 1 will be compatible regardless of which frequency is assigned, so
+    # we must use radio 1 for the intra-flight radio.
+    intra_flight_radio_index: 1
+    inter_flight_radio_index: 2
```
Raffson: The radios should be left out since there's no custom radios in the pydcs-extension, so this will likely lead to problems during mission generation if my guess is right 😅

Raffson (approved):

PR #660, opened 2026-01-27
Fix state writing on mission end event

red-one1:
Changes the order that saving is done on mission end.
This fixed my "mission_ended:true" bug.


Raffson (approved):

Raffson:
Looks pretty straightforward to me 👍

PR #668, opened 2026-02-02
Escorts are now kept on a leash.

red-one1:
Escorts engagement range is now from the bombers, not from themselves.
This should stop them from going hunting and leaving the bombers to die, when there's a lot of hostiles up.

### DIFF
--- a/game/missiongenerator/aircraft/flightgroupconfigurator.py
+++ b/game/missiongenerator/aircraft/flightgroupconfigurator.py
@@ -28,6 +28,7 @@ from game.radio.tacan import (
     OutOfTacanChannelsError,
 )
 from game.runways import RunwayData
+from game.missiongenerator.missiondata import EscortInfo
 from game.squadrons import Pilot
 from .aircraftbehavior import AircraftBehavior
 from .aircraftpainter import AircraftPainter
@@ -145,7 +146,7 @@ class FlightGroupConfigurator:
             self.flight.flight_plan.waypoints,
         )
 
-        return FlightData(
+        flight_data = FlightData(
             package=self.flight.package,
             aircraft_type=self.flight.unit_type,
             squadron=self.flight.squadron,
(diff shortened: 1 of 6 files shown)


game/missiongenerator/aircraft/flightgroupconfigurator.py
```diff
@@ -169,6 +170,42 @@ def configure(self) -> FlightData:
...
+            FlightType.SEAD_ESCORT,
+        ]:
+            return
+
+        if self.flight.package.primary_flight is None:
+            return
+
+        escort_group_id = self.flight.group_id
+        escorted_group_id = self.flight.package.primary_flight.group_id
+        if escort_group_id <= 0 or escorted_group_id <= 0:
```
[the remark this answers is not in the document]
Raffson: Can be ignored cause this would only occur with Pretense, in which case there's no real "escorts" so no need to log a warning since the spawner-class should ensure the ID was set

game/missiongenerator/luagenerator.py
```diff
@@ -273,10 +273,74 @@ def generate_plugin_data(self) -> None:
...
+            escort_item.add_key_value(
+                "engagementRangeMeters", str(escort.engagement_range_meters)
+            )
+
         trigger = TriggerStart(comment="Set DCS Retribution data")
         trigger.add_action(DoScript(String(lua_data.create_operations_lua())))
         self.mission.triggerrules.triggers.append(trigger)
 
+        if self.mission_data.escorts:
+            escort_lua = """
```
Raffson: This block of Lua-code should probably be put into the base-script 🤔 
-> https://github.com/dcs-retribution/dcs-retribution/blob/dev/resources/plugins/base/dcs_retribution.lua

game/missiongenerator/luagenerator.py
```diff
@@ -273,10 +273,74 @@ def generate_plugin_data(self) -> None:
...
+                    if not dcsRetribution or not dcsRetribution.Escorts then
+                        return nil
+                    end
+
+                    for _, pair in pairs(dcsRetribution.Escorts) do
+                        local escort_group = get_group(tonumber(pair.escortGroupId))
+                        local escorted_group = get_group(tonumber(pair.escortedGroupId))
+                        if escort_group and escorted_group then
+                            local escort_unit = escort_group:getUnit(1)
+                            local escorted_unit = escorted_group:getUnit(1)
```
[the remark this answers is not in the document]
red-one1: Done

red-one1 (commented):
@Raffson please re-review!

resources/plugins/base/dcs_retribution.lua
```diff
@@ -196,3 +196,63 @@ mist.addEventHandler(onEvent)
...
+            if escort_unit and escorted_unit then
+                local escort_pos = escort_unit:getPoint()
+                local escorted_pos = escorted_unit:getPoint()
+                local dx = escort_pos.x - escorted_pos.x
+                local dz = escort_pos.z - escorted_pos.z
+                local distance = math.sqrt(dx * dx + dz * dz)
+
+                local max_dist = tonumber(pair.engagementRangeMeters) or 0
+                if max_dist > 0 and distance > max_dist then
+                    escort_leash_set_roe(escort_group, AI.Option.Air.val.ROE.WEAPON_HOLD)
```
Raffson: Perhaps this too strict? Perhaps RETURN_FIRE makes more sense here? 🤔

Raffson: Went ahead and switched to RETURN_FIRE, because if the escort is under attack, it will basically surrender 😅

red-one1: Makes sense to me

Raffson (approved):

PR #674, opened 2026-02-06
Revamp SEAD/DEAD planning

prokop7:
Current implementation of DEAD/SEAD produces suboptimal packages
Issue is: the planner prioritizes SAMs that are not relevant to the theater, for example flying 150 km to destroy an SA-13 deep in the backline while completely ignoring an SA-5 located near the border. 

The solution this PR has:
SEAD/DEAD planner now targets the most threatening SAM sites closest to friendly control points. It calculates range from each SAM to the closest CP with consideration of a SAMs range

### DIFF
--- a/game/commander/tasks/compound/degradeiads.py
+++ b/game/commander/tasks/compound/degradeiads.py
@@ -1,5 +1,5 @@
 import logging
-from collections.abc import Iterator
+from collections.abc import Iterator, Sequence
 from typing import Union
 
 from game.commander.tasks.primitive.antiship import PlanAntiShip
@@ -7,8 +7,9 @@ from game.commander.tasks.primitive.dead import PlanDead
 from game.commander.theaterstate import TheaterState
 from game.data.groups import GroupTask
 from game.htn import CompoundTask, Method
+from game.theater.missiontarget import MissionTarget
 from game.theater.theatergroundobject import IadsGroundObject, NavalGroundObject
-from game.utils import meters
+from game.utils import Distance
 
 
 class DegradeIads(CompoundTask[TheaterState]):
@@ -18,6 +19,7 @@ class DegradeIads(CompoundTask[TheaterState]):
     def each_valid_method(self, state: TheaterState) -> Iterator[Method[TheaterState]]:
         # Get all candidate targets in priority order
         candidates = []
+        settings = state.context.settings
 
         # Priority 1: Strategic threats (based on threat circle proximity)
         strategically_threatening_sams = self._assess_strategic_threats(
@@ -27,7 +29,13 @@ class DegradeIads(CompoundTask[TheaterState]):
             candidates.append(("strategic", sam))
 
         # Priority 2: Systems actively threatening planned missions
-        prioritized_threats = self._prioritize_threatening_systems(state.threatening_air_defenses)
+        prioritized_threats = self._prioritize_threatening_systems(
+            [
+                defense
+                for defense in state.threatening_air_defenses
+                if isinstance(defense, IadsGroundObject)
+            ]
+        )
         for sam in prioritized_threats:
             # Only add if not already in candidates
             if not any(c[1] == sam for c in candidates):
@@ -35,10 +43,11 @@ class DegradeIads(CompoundTask[TheaterState]):
 
         # Priority 3: High-value detection systems (EWRs)
         high_value_detectors = [
-            d for d in state.detecting_air_defenses
+            d
+            for d in state.detecting_air_defenses
             if d.task == GroupTask.EARLY_WARNING_RADAR
         ]
-        for ewr in high_value_detectors[:2]:  # Top 2 EWRs
+        for ewr in high_value_detectors[:1]:
             if not any(c[1] == ewr for c in candidates):
                 candidates.append(("detector", ewr))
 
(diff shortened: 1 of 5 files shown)


Raffson:
There's a little more to SEAD/DEAD, i.e. it should focus primarily on IADS which are preventing other flights from hitting their target, but there's definitely room for improvement in that sense that it could consider an SA-13 that's right next to a target more important than an SA-10 which is 30NM further. I _think_ the place you want to look at is https://github.com/dcs-retribution/dcs-retribution/blob/dev/game/commander/tasks/packageplanningtask.py#L163 🤔

prokop7:
I'll investigate why sometimes I get only sead in 150nm without any other packages that this sead is supporting
I thought about getting deep strike combined with sead, or use sead to clear the way for barcap or any other activities 

Nevertheless, I see some improvements on dead packages with this version. Gonna get more testing rather than coding :)

Btw, If you allow it to be built at ci/cd, community would be able to test it on their own too

Raffson:
Feel free to share a save file once you found such a case, then we could have a look as well 😉

prokop7:
Can't find save with my case, but here is another that shows that something is wrong with SEAD/DEAD
Clean start, Dust to Dust campaign, NO SEAD at all :)
Nobody would fly those BAI without cleaning HAWK and SA11 first
image

Compare after updates:
image

Yes, SA5 above isn't necessary to clean, this requires tuning...

prokop7:
We can go into discord if you wish :)
There is a lot easier to share with you missions and comparison

prokop7:
Black Sea. Clean Start

Before
image

After
image

Raffson:
These results do look promising 👌
I'll try to properly review tomorrow behind a computer, cause currently I'm reviewing everything on my phone 😅

game/commander/tasks/compound/nextaction.py
```diff
@@ -29,10 +29,10 @@ def each_valid_method(self, state: TheaterState) -> Iterator[Method[TheaterState
         yield [ProtectAirSpace()]
         yield [DefendBases()]
         yield [InterdictReinforcements()]
+        yield [DegradeIads()]
```
Raffson, line 32: This should be reverted because otherwise `threatening_air_defenses` won't have been fully populated. It's during planning that `threatening_air_defenses` is filled up to figure out which IADS are supposed to be targeted.

game/commander/tasks/compound/degradeiads.py
```diff
@@ -1,33 +1,249 @@
...
+        Focus on high-value threats: LORAD > MERAD > SHORAD.
+        AAA is only targeted if it's one of the very few threats.
+        """
+        if not threatening:
+            return []
+
+        # Categorize by threat tier
+        lorad_threats = [t for t in threatening if t.task == GroupTask.LORAD]
+        merad_threats = [t for t in threatening if t.task == GroupTask.MERAD]
+        shorad_threats = [t for t in threatening if t.task == GroupTask.SHORAD]
```
Raffson, line 73: Unused variable? 🤔

Raffson (changes requested):
I'm still not convinced, cause when I take a closer look at your before & after screenshots, I notice a lot less flights are planned against targets around that first CP, which isn't really desirable. Also, the majority of this code seems to do what used to be done already before, aside from switching up the order of things maybe, because active threats against planned missions should take priority over so called "strategic threats", which is more or less the same as what `prioritized_air_defenses` was about, with the main difference being conciseness... The only upside I see here is the configurability of things, but that could be introduced in a much cleaner way with much less code 😅

game/commander/tasks/compound/degradeiads.py
prokop7: Tried different things, forget to remove 👍

game/commander/tasks/compound/nextaction.py
prokop7: I would suggest to split this class into DegradeIads and AttackSamSites
DegradeIads could only populate threatening_air_defenses in the same way as other actions do and AttackSamSites will "execute" those threatening_air_defenses.

By lowering everything where DegradeIads was, you will not have impact on "strategic" SAM sites because threatening_air_defenses  is overpopulated already

What do you think?

prokop7:
image
Some feedback
Also, tried to play few campaigns and I've encountered some better results with LORAD on map and mixed results when only MERAD/SHORAD are present

game/commander/tasks/compound/nextaction.py
Raffson: Aside from the fact that just like BAI, the majority of this code isn't sitting in the right place, I still think you're chasing the wrong thing in order to get the desired effect. If what you say is true, then the solution would be to limit the number of flights that get planned, which should automatically limit the number of threatening IADS. Additionally the targets against which something is planned (whether that be BAI, Strike, etc., doesn't really matter) should be ordered according to their priority (distance relative to the closest friendly CP is usually ok, though it might make more sense to take an average of X closest CPs, where X should be configurable).
Either way, anything that's relying on `threatening_air_defenses` to plan missions should be last in the list here, because that's how it was architecturally set up. That list gets filled up whenever the preconditions are checked, so no matter how you turn those tables, your changes would not be optimal due to the fact that your still relying on the `threatening_air_defenses`, which will be an incomplete list.
If I were you, I'd start fresh, and just focus on ordering `threatening_air_defenses` according to its distance to the closest friendly CP, or as I mentioned earlier, a set of X closest CPs to get an average, and see whether that has a positive effect. Bottom line, try to keep it simple because things are already complicated enough as they currently stand 😉

PR #675, opened 2026-02-06
Revamp mission scheduling

prokop7:
Problem:
Scheduling of long-duration flights. For example, a 50-minute flight with a desired mission duration of 60 minutes can be scheduled with a TOT of 1:30+ after mission start.

This PR tries to solve this problem. Longer missions spawn earlier, making them more likely to fit within the desired mission duration window.

### DIFF
--- a/game/commander/missionscheduler.py
+++ b/game/commander/missionscheduler.py
@@ -20,6 +20,47 @@ class MissionScheduler:
         self.coalition = coalition
         self.desired_mission_length = desired_mission_length
 
+    def _cluster_packages_by_location(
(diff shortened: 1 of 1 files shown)


Raffson:
I think this a non-issue, it's rather bad documentation of the mission-duration setting, which is basically a value to indicate when the latest possible start-time instead of the actual duration of the mission. In general, it's quite difficult to put an estimate on mission duration cause it all depends on how far the target is going to be 😅

prokop7:
I agree with you, closing this PR

PR #676, opened 2026-02-06
Revamp BARCAP planning

prokop7:
Issue with current BARCAP planning is too few control points have CAP coverage. The current settings can create BARCAPs randomly, which is not ideal. OPFOR air presence is almost non-existent.

This PR adds additional BARCAPs for the closest to enemy control points. Also, BARCAPs are scheduled ASAP to provide air superiority for other packages and defend crucial targets

### DIFF
--- a/game/commander/tasks/packageplanningtask.py
+++ b/game/commander/tasks/packageplanningtask.py
@@ -111,14 +111,8 @@ class PackagePlanningTask(TheaterCommanderTask, Generic[MissionTargetT]):
             state.context.settings,
         )
         with state.context.tracer.trace(f"{color} {self.flights[0].task} planning"):
-            asap = False
-            if (
-                not state.context.coalition.ato.has_awacs_package
-                and FlightType.AEWC in [f.task for f in self.flights]
-            ):
-                asap = True
             self.package = fulfiller.plan_mission(
-                ProposedMission(self.target, self.flights, asap=asap),
+                ProposedMission(self.target, self.flights, asap=self.asap),
                 self.purchase_multiplier,
                 state.context.now,
                 state.context.tracer,
--- a/game/commander/tasks/primitive/barcap.py
+++ b/game/commander/tasks/primitive/barcap.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from logging import getLogger
 
 from game.ato.flighttype import FlightType
 from game.commander.tasks.packageplanningtask import PackagePlanningTask
@@ -13,8 +14,8 @@ class PlanBarcap(PackagePlanningTask[ControlPoint]):
     max_orders: int
 
     def preconditions_met(self, state: TheaterState) -> bool:
-        if not state.barcaps_needed[self.target]:
-            return False
+        if not state.barcaps_needed.get(self.target, 0):
+            state.barcaps_needed[self.target] = 1
         return super().preconditions_met(state)
 
     def apply_effects(self, state: TheaterState) -> None:
@@ -28,3 +29,7 @@ class PlanBarcap(PackagePlanningTask[ControlPoint]):
     @property
     def purchase_multiplier(self) -> int:
         return self.max_orders
+
+    @property
+    def asap(self) -> bool:
+        return True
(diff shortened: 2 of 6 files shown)


game/commander/tasks/primitive/barcap.py
```diff
@@ -28,3 +29,7 @@ def propose_flights(self) -> None:
     @property
     def purchase_multiplier(self) -> int:
         return self.max_orders
+
+    @property
+    def asap(self) -> bool:
+        return True
```
Raffson, line 35: What's the meaning of this? Did CoPilot do this? 😅

game/commander/tasks/packageplanningtask.py
```diff
@@ -111,14 +111,8 @@ def fulfill_mission(self, state: TheaterState) -> bool:
             state.context.settings,
         )
         with state.context.tracer.trace(f"{color} {self.flights[0].task} planning"):
-            asap = False
-            if (
-                not state.context.coalition.ato.has_awacs_package
-                and FlightType.AEWC in [f.task for f in self.flights]
-            ):
-                asap = True
```
Raffson, line 119: This seems to have gone lost, or did I miss something? 😅

game/commander/theaterstate.py
```diff
@@ -190,12 +190,43 @@ def from_game(
...
+                min_distance = min(
+                    cp.position.distance_to_point(enemy.position)
+                    for enemy in enemy_cps
+                )
+                cp_distances.append((min_distance, cp))
+            
+            # Sort by distance and take the 3 closest
+            cp_distances.sort(key=lambda x: x[0])
+            for _, cp in cp_distances[:3]:
+                barcaps_needed[cp] = 1  # Only 1 round for supplemental coverage
```
Raffson: This looks incorrect, the number of barcaps depends on the mission duration divided by the duration of a barcap 🤔

Raffson (changes requested):

game/commander/tasks/primitive/barcap.py
prokop7: IDK, I hope you tell me :)
This is from AEWC class 

I've analysed that you have this property and it was never used, though I made it useful

game/commander/tasks/packageplanningtask.py
prokop7: That was the logic of how to make aewc packages asap, now it relies on property within class, rather than hardcode here

game/commander/theaterstate.py
prokop7: Yeah, but I didn't want to spend too much aircraft on additional BARCAPs.
I'll fix it though

game/commander/tasks/primitive/barcap.py
prokop7: https://github.com/dcs-retribution/dcs-retribution/blob/b901b78e7b775006570a41a90c6b56db34057dea/game/commander/tasks/primitive/aewc.py#L31-L34

Raffson: I'm afraid this isn't going to result in the same behavior, because now ASAP is always true for BARCAPs, which shouldn't be the case. This forces ASAP to be true when there's no AWACS package planned yet, and the current package has an AWACS flight in it, unless I'm missing something 😅

game/commander/theaterstate.py
Raffson: Given that both parameters are configurable, this could potentially lead to confused users 😅

game/commander/tasks/primitive/barcap.py
prokop7: I think I understand, but tell me if I’m wrong.
My idea was to launch all BARCAPs ASAP, in the same way AWACS missions are launched.
However, this would break consecutive BARCAPs and AWACS missions, since they would all be scheduled ASAP, right?

prokop7:
Made params configurable and fixed number of rounds

game/commander/tasks/primitive/barcap.py
Raffson: Indeed, and so the condition in packageplanningtask which you removed under the flag "useless" was in fact to ensure that the first AEWC package does in fact start up asap so that they have the earliest possible TOT 😅
Granted, the TOT also depends very much on the departure location relative to the "TOT-waypoint", but that's a different story 😜

prokop7: Well, I'll revert my logic, and I will look into how to make a barcap asap

PR #677, opened 2026-02-06
Add reducing economy strength planning

prokop7:
Current version of planner attacks building randomly. I've added a meaning to those attacks. Now they prioritizing highest income buildings which are out of threatening zones and relatively close.

Although, algorithm is not ideal and needs a proper tuning. Often more deep strikes are preferrable, which current version will probably skip

### DIFF
--- a/game/commander/tasks/compound/attackinfrastructure.py
+++ b/game/commander/tasks/compound/attackinfrastructure.py
@@ -1,0 +1,91 @@
+from __future__ import annotations
+
+from collections.abc import Iterator
+from dataclasses import dataclass
+
+from game.commander.tasks.primitive.strike import PlanStrike
+from game.commander.theaterstate import TheaterState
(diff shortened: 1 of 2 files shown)


Starfire13:
Yeah, prioritising deep strikes would be good, as it incentivises clearing the way for such deep strikes by taking down SAMs protecting such targets.

Raffson (commented):
To me it looks like this is very similar to AttackBuildings, aside that you're taking some economic parameters into account which AttackBuildings doesn't. It's based on distance in fact, not random as you suggest in the description of the PR 😉

prokop7:
Thank you for your review!
I had seen similarities, though about modifying existed
As Starfire said, I would make it to prioritize deep strikes instead of close one

Raffson:
That said, I think this adjustment should probably be merged into AttackBuildings straight away 🤔

PR #678, opened 2026-02-06
Revamp BAI planning

prokop7:
Targets are often not relevant to the theater, resulting in an almost random selection of target groups across the map. Often BAI targets are too far away (100nm+ on big maps)

This solves this problem, BAI now targets groups near the closest control points.

However, this need a lot of tuning to get it right. Community help would be appreciated

### DIFF
--- a/game/commander/tasks/compound/attackbattlepositions.py
+++ b/game/commander/tasks/compound/attackbattlepositions.py
@@ -6,69 +6,66 @@ from game.commander.tasks.primitive.bai import PlanBai
 from game.commander.theaterstate import TheaterState
 from game.htn import CompoundTask, Method
 
-# Maximum number of BAI packages to plan
-MAX_BAI_PACKAGES = 4
-# Maximum number of battle positions per enemy CP
-MAX_POSITIONS_PER_CP = 2
-# Maximum distance to consider enemy CP for BAI (in meters)
-MAX_BAI_DISTANCE = 200000  # 200km
-
 
 class AttackBattlePositions(CompoundTask[TheaterState]):
     def each_valid_method(self, state: TheaterState) -> Iterator[Method[TheaterState]]:
         """Plan BAI missions against nearest enemy control points.
-        
+
         Prioritizes enemy CPs that are closest to friendly CPs.
         Limits total BAI packages to prevent spending all aircraft on ground attack.
         """
         # Get friendly control points for distance calculation
-        friendly_cps = list(state.context.theater.control_points_for(
-            state.context.coalition.player
-        ))
-        
+        friendly_cps = list(
+            state.context.theater.control_points_for(state.context.coalition.player)
+        )
+
         if not friendly_cps:
             return
-        
+
+        settings = state.context.settings
+
         # Score and sort enemy CPs by distance to nearest friendly CP
         enemy_cp_scores = []
         for enemy_cp, battle_positions in state.enemy_battle_positions.items():
             # Skip if no battle positions
             if not list(battle_positions.in_priority_order):
                 continue
-            
+
             # Find distance to closest friendly CP
             min_distance = min(enemy_cp.distance_to(fcp) for fcp in friendly_cps)
-            
+
             # Filter out CPs that are too far away
-            if min_distance > MAX_BAI_DISTANCE:
+            if min_distance > settings.bai_max_distance_meters:
                 continue
-            
+
             # Lower distance = higher priority (negate for sorting)
             enemy_cp_scores.append((min_distance, enemy_cp, battle_positions))
-        
+
         # Sort by distance (closest first)
         enemy_cp_scores.sort(key=lambda x: x[0])
-        
-        # Yield BAI targets from closest CPs, limited to MAX_BAI_PACKAGES total
+
+        # Yield BAI targets from closest CPs, limited to configured total
         packages_yielded = 0
         for distance, enemy_cp, battle_positions in enemy_cp_scores:
-            if packages_yielded >= MAX_BAI_PACKAGES:
+            if packages_yielded >= settings.bai_max_packages:
                 break
-            
+
             # Yield top battle positions from this CP
             positions_from_cp = 0
             for battle_position in battle_positions.in_priority_order:
-                if positions_from_cp >= MAX_POSITIONS_PER_CP:
+                if positions_from_cp >= settings.bai_max_positions_per_cp:
                     break
-                
+
                 yield [PlanBai(battle_position)]
                 packages_yielded += 1
                 positions_from_cp += 1
-                
-                if packages_yielded >= MAX_BAI_PACKAGES:
+
+                if packages_yielded >= settings.bai_max_packages:
                     break
-        
+
         # Plan armed recon against highest priority CPs
-        for cp in state.control_point_priority_queue[:2]:
+        for cp in state.control_point_priority_queue[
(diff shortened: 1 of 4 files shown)


game/commander/tasks/compound/attackbattlepositions.py
```diff
@@ -1,17 +1,74 @@
...
+MAX_BAI_PACKAGES = 4
+# Maximum number of battle positions per enemy CP
+MAX_POSITIONS_PER_CP = 2
+# Maximum distance to consider enemy CP for BAI (in meters)
+MAX_BAI_DISTANCE = 200000  # 200km
+
 
 class AttackBattlePositions(CompoundTask[TheaterState]):
     def each_valid_method(self, state: TheaterState) -> Iterator[Method[TheaterState]]:
-        for battle_positions in state.enemy_battle_positions.values():
```
Raffson, line 11: enemy_battle_positions has a pretty specific meaning, targeting TGOs that are blocking a capture, so to me it looks strange to see this disappear 😅

Raffson (commented):
Also, can you provide some examples of "irrelevant" BAI target? 😅

prokop7:
> Also, can you provide some examples of "irrelevant" BAI target? 😅

Here is an example of such BAI. I don't think these BAI are relevent (i've removed all other packages, there was no DEAD, only BARCAP and 1xStrike)
Frontline is 8+ control points away from those targets
image

game/commander/tasks/compound/attackbattlepositions.py
prokop7: It didn't. Here it is on line 34

Raffson:
How many CPs can you find where you have TGOs within
3000 meters (or 2500, can't remember which one it is) of tje CP?

Raffson:
Also, take a look at the following method: https://github.com/dcs-retribution/dcs-retribution/blob/dev/game/commander/battlepositions.py#L31

prokop7:
Here is same state of the game with all mine fixes (quite less BAI because of DEAD packages)
Again, cleaned up everything except BAI
image

About 200km, probably irrelevant. But is is just limiter for BAI, it can be 50km and then without enemy's ground units within this range no BAI will be created. Otherwise you fill fly 100km+ on A10 :)

prokop7:
> Also, take a look at the following method: https://github.com/dcs-retribution/dcs-retribution/blob/dev/game/commander/battlepositions.py#L31

I've seen it
Still, looking at those targets from first exmaple, they are not blocking capture of points. Otherwise you can say it about anything that has "enemy" tag on it :)

Raffson:
> > Also, take a look at the following method: https://github.com/dcs-retribution/dcs-retribution/blob/dev/game/commander/battlepositions.py#L31
> 
> 
> 
> I've seen it
> 
> Still, looking at those targets from first exmaple, they are not blocking capture of points. Otherwise you can say it about anything that has "enemy" tag on it :)

I see your point, but I'd argue that your code is not sitting in the right place. I _think_ offloading the logic to BattlePositions makes more sense, though I'd have to take a closer look tomorrow 😅

game/commander/tasks/compound/attackbattlepositions.py
Raffson: Right, my bad 😅 👍

game/commander/tasks/compound/attackbattlepositions.py
```diff
@@ -8,10 +9,63 @@
...
+        Prioritizes enemy CPs that are closest to friendly CPs.
+        Limits total BAI packages to prevent spending all aircraft on ground attack.
+        """
+        # Get friendly control points for distance calculation
+        friendly_cps = list(
+            state.context.theater.control_points_for(state.context.coalition.player)
+        )
+
+        if not friendly_cps:
+            return
```
Raffson, line 23: This is redundant afaik, the game would've been lost and the code shouldn't get to this point. Correct me if I'm wrong, but this is all AI generated, right? 😅

game/commander/tasks/compound/attackbattlepositions.py
Raffson, line 1: All of this code doesn't belong here. Keep in mind, `each_valid_method` is a generator, so having this many conditions and what not is far from ideal, especially from a performance point of view 😅

I'd look into offloading this into `BattlePositions` itself, or somehow do the filtering in `TheaterState` 🤔

game/commander/tasks/compound/attackbattlepositions.py
```diff
@@ -8,10 +9,63 @@
...
-        for cp in state.control_point_priority_queue[:2]:
+                packages_yielded += 1
+                positions_from_cp += 1
+
+                if packages_yielded >= settings.bai_max_packages:
+                    break
+
+        # Plan armed recon against highest priority CPs
+        for cp in state.control_point_priority_queue[
+            : settings.bai_armed_recon_target_count
```
Raffson, line 68: This is technically unrelated to BAI, so maybe the setting-name should be adjusted 🤔

Raffson (changes requested):

game/commander/tasks/compound/attackbattlepositions.py
prokop7: Kinda so, I've tried to fix many things it got wrong, however it isn't perfect :)

prokop7: Yet you doesn't know when this code is invoked, better to use some pre-conditions to keep it safe from exploding

Raffson: The danger with that is the fact that it produces way too much code than what is actually needed, but I can definitely appreciate the attempt 😅

game/commander/tasks/compound/attackbattlepositions.py
prokop7: I'm looking at how this can be at BattlePositions  and can't see it. BattlePositions strictly tied to ControlPoints which should be AirAssaulted and... Taking BAI targets from there lead us to this PR :D
Filtering BattlePositions at TheaterState would lead to similar results if there is no relevant BP at the begging

Performance is concerned, I know, though I can't see substantial difference in time between vanila and this

prokop7: I could move out this code to TheaterState and provide additional list of bai targets, but this would be no much better than current version (except performance)

prokop7: I’ve tried this, but it makes things messy, and TheaterState starts turning into a superclass, which it shouldn’t.
In my opinion, either enemy_battle_positions or BAI targets should be kept—having both doesn’t work well together.

Raffson: Essentially what you're doing is applying a filter, so this is more something that relates to ObjectiveFinder. Arguably some of the adjustments I've made over the years weren't always in the right place either, but in this case I think the code should probably sit in its own class and serve as a helper for ObjectiveFinder in order to prioritize targets in a better way. Does that make sense? 😅

prokop7: Sounds about right, I'll look into it
I don't really like current implementation, even if I like the results :)

prokop7: Sounds about right, I'll look into it
I don't really like current implementation, even if I like the results :)

PR #679, opened 2026-02-07
Add ground units repairs

prokop7:
The AI can plan repairs for ground units (currently only SAMs are supported).
It identifies the most valuable SAM sites and prioritizes repairing them. The algorithm prioritizes search and tracking radars to make SAM systems operational. The repair budget is shared with ground units and limited to 40% of the it.
All ground objects are purchased or repaired with a configurable delay rather than instantly.

There are 2 new configs: 
- Automate ground object repairs:  If enabled, AI can spend budget to repair destroyed ground object units such as SAMs and EWRs.
- Ground object repair turns: Turns required for repaired ground object units to return to service. Set to 0 for instant repairs.

Here is how such units are displayed
image

### DIFF
--- a/game/armedforces/forcegroup.py
+++ b/game/armedforces/forcegroup.py
@@ -220,6 +220,17 @@ class ForceGroup:
     ) -> TheaterGroundObject:
         """Create a TheaterGroundObject for the given template"""
         go = layout.create_ground_object(name, location, control_point, task)
+        required_classes = set()
+        for unit_group in layout.all_unit_groups:
+            if unit_group.optional:
+                continue
+            required_classes.update(unit_group.unit_classes)
+            for unit_type in unit_group.unit_types:
+                if issubclass(unit_type, VehicleType):
+                    required_classes.add(
+                        next(GroundUnitType.for_dcs_type(unit_type)).unit_class
+                    )
+        go.required_unit_classes = required_classes
         # Generate all groups using the randomization if it defined
         for tgo_group in layout.groups:
             for unit_group in tgo_group.unit_groups:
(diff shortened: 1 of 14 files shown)


Starfire13:
Might be good to make the percentage of ground unit $$$ that can be used for repairs configurable.

prokop7:
Yeah, I thought about it and I need your help on it. 
Current setting is 40% of frontline units budget and settings may confuse users. If frontline units budget is 50%, then this would result in 20% only, however it is shared. AI may not spend on SAM and therefore spend more on frontline units.

How should I do? Make it 20% of total budget? Make it 40% of frontline units?

Also, I'm working on repairing buildings and it will have budget too...

Raffson:
Arguably those ratios should all be configurable, probably best under campaign management I guess 😅

prokop7:
Yeah, wouldn't it be too much for users? :)

How about hiding these settings under a collapsible spoiler with advanced settings and not overwhelming all users?

Raffson:
> Yeah, wouldn't it be too much for users? :)
> 
> How about hiding these settings under a collapsible spoiler with advanced settings and not overwhelming all users?

Not a bad idea, though I wonder how much of a hassle this will turn out to be with PyQt 😅

prokop7:
Or I could make a pile of advanced options and separate them through the left menu. Call it "Advanced campaign management"
That would be much easier

juanjux:
This is a cool feature that could make the campaigns more balanced (since currently only the player can repair stuff), hope somebody with write permissions could review soon (ping @Druss99 and @Starfire13)

juanjux:
Would be cool to also allow the AI to repair tanks/trucks/BMPs on base armor groups.(the same kind that the player can repair when DEAD). However, could be perfectly be another PR.

Druss99:
I think this needs a fresh rebase and it looks like there are some random unrelated changes throughout probably from rebasing previously

PR #680, opened 2026-02-07
Add ground buildings repairs

prokop7:
The AI can plan repairs for buildings. Additionally, users can now repair buildings on demand.
It identifies the most valuable buildings that are farthest from the opponent. An exception is ammo depots, which must be closest to the frontline. The potential income value of a building is also taken into consideration.
Building repairs have a configurable and substantial delay, with a default of 4 turns. The repair cost is calculated as income_value × 4, subject to change. There is also a new repair budget option with a default value of 15% of the total budget.

The current implementation does not support refunding or canceling repairs, so please keep that in mind.

Examples:
image
image
image

### DIFF
--- a/game/ato/flight.py
+++ b/game/ato/flight.py
@@ -129,7 +129,7 @@ class Flight(
         # altitude offset for planes
         offset_factor = self.coalition.game.settings.max_plane_altitude_offset
         offset_factor = random.randint(0, offset_factor)
-        self.plane_altitude_offset = 1000 * offset_factor * random.choice([-1, 1])
+        self.plane_altitude_offset: int = 1000 * offset_factor * random.choice([-1, 1])
 
     @property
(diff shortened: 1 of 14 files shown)


Starfire13:
4 turns is the same as runway repairs, which seems terribly short. Considering each turn is only a couple hours apart, that would mean a factory can be rebuilt (and all its machinery replaced) from rubble within a day or two.

prokop7:
Sure, it's a game :)
Turns are configurable. You can buy the whole squadron in one turn, you can capture 4 control points, buy 20 MBT and many other things. Why not repair some holes in the roof of a factory?

juanjux:
Cool feature! Regarding the discussion, I agree with @prokop7 that a default of 4 turns should be fine, considering how much can happen in 1 turn in this campaign, also, if somebody doesn't like it, he made it as an option.

PR #683, opened 2026-02-07
add weapon restriction overrides per faction

Druss99:
Adds weapon restriction overrides on a faction level. Cherry picked from liberation.


Raffson (approved):
LGTM 👍

PR #684, opened 2026-02-07
Simplify fast forward settings

Druss99:
This PR simplifies fast forward settings and introduces the ability to skip combat instead of resolving.


Raffson (approved):
Architecturally it all makes sense, and I can't immediately spot something that's wrong so let's go ahead and merge this one 🤞

PR #693, opened 2026-02-13
Updating Vietnam War Vessels to v3.0.0

Biggus22:
This PR is to update Vietnam War Vessels to v3.0.0.

https://github.com/tspindler-cms/tetet-vwv/releases/tag/VWV_3.0.0

A couple of relevant factions needed some minor changes to reflect some changes to units within the mod.


resources/units/ships/vap_us_seafloat.yaml
```diff
@@ -1,4 +1,4 @@
 class: Boat
 price: 0
 variants:
-  VAP - US Sea Float Barge: null
```
Druss99, line 4: Removing this variant will break save compatibility unless it is migrated. Let's leave it in the file and just add the other one that is more nicely named

resources/units/ground_units/vap_zis_150_aa.yaml
```diff
@@ -3,4 +3,4 @@ description:
 price: 7
 role: Self-Propelled Anti-Aircraft Gun
 variants:
-  VAP VC Zis 150 AAA: {}
```
Druss99, line 6: See comment on other unit about removing variants

Druss99 (changes requested):

resources/units/ships/vap_us_seafloat.yaml
Biggus22: Hopefully this is fixed now.  That's a killer of a difference in naming to spot.

resources/units/ground_units/vap_zis_150_aa.yaml
Biggus22: Hopefully this one is fixed too.

pydcs_extensions/vietnamwarvessels/vietnamwarvessels.py
```diff
@@ -2577,3 +2487,401 @@ class Pylon4:
...
+@vehiclemod
+class vap_m35_gun_truck(unittype.VehicleType):
+    id = "vap_m35_gun_truck"
+    name = "VAP US M35 Gun Truck"
+    detection_range = 0
+    threat_range = 4500
+    air_weapon_dist = 4500
+
+
+class vap_zis_150_aa(unittype.VehicleType):
```
Druss99, line 2568: quite a few of the vehicles here aren't marked as mods with the @vehicle_mod tag, is that intentional?

Druss99:
My only last question is that I thought this version was supposed to combine the VWV and VAP mods into one pack. If that's the case I would've expected more units but maybe I am missing something? Also I will note that long term this will be quite an undertaking since all the units from one mod will need to be migrated to the other but I think that can wait for now if users can enable both without issue.

Druss99 (changes requested):

pydcs_extensions/vietnamwarvessels/vietnamwarvessels.py
Biggus22: Should be fixed now, not intentional, just inattentive.
I've also moved the new statics up to the block containing older statics.  I think in the future I'll just leave the new statics accumulating at the bottom on the file with a comment pointing readers up to the original block.

Biggus22:
> My only last question is that I thought this version was supposed to combine the VWV and VAP mods into one pack. If that's the case I would've expected more units but maybe I am missing something? Also I will note that long term this will be quite an undertaking since all the units from one mod will need to be migrated to the other but I think that can wait for now if users can enable both without issue.

Maybe they're doing in piecemeal.  I'm just going with what the export sends out.  Will see what happens next.

Druss99:
> Maybe they're doing in piecemeal.  I'm just going with what the export sends out.  Will see what happens next.

I see. My confusion here is that there are a lot of changes to the pydcs extension but only a few unit yaml changes. Is the intention that most of the units aren't needed in a retribution context OR do they already have unit yamls from the VAP mod? If they already have unit yamls from the other mod then I think we need to add lines in faction.py to remove the units if this mod isn't enabled.

Hmm but that gets messy because technically you only need VAP or VWV enabled for the migrated units, not both...

I think we probably need to hold off on this until they fully combine the mods so we can migrate all the VAP units over and close it out, otherwise we don't have a good way to handle units that are provided by 2 different mods.

Biggus22:
> > Maybe they're doing in piecemeal.  I'm just going with what the export sends out.  Will see what happens next.
> 
> I see. My confusion here is that there are a lot of changes to the pydcs extension but only a few unit yaml changes. Is the intention that most of the units aren't needed in a retribution context OR do they already have unit yamls from the VAP mod? If they already have unit yamls from the other mod then I think we need to add lines in faction.py to remove the units if this mod isn't enabled.
> 
> Hmm but that gets messy because technically you only need VAP or VWV enabled for the migrated units, not both...
> 
> I think we probably need to hold off on this until they fully combine the mods so we can migrate all the VAP units over and close it out, otherwise we don't have a good way to handle units that are provided by 2 different mods.

I hadn't actually counted the VAP units vs the VWV units, but could see lots that didn't have yamls so I'd assumed that there was still quite a bit missing.  Turns out they're all there, and the units that don't have a yaml don't really need a yaml.  They are ammoboxes, loudspeakers, radios, that kind of thing.

There's also a fair few changes to existing units within the update.

It's probably the most confusing mod to try to support, because not only is there the incorporation of the VAP stuff, but also there are assets with the prefix VAP that aren't in the current OH6 assets pack that have been added to VWV in this update and they have kept their VAP prefixes.  I've updated faction.py to hopefully handle the cases where one mod might be active and the other not.

Druss99 (approved):

PR #698, opened 2026-02-22
More state.json fixes

### LINKED ISSUE #697: More state.json issues
### Affected versions

Test build

### Test/Development build

2406

### Description

>Likely an idiotic question, but just getting back into this and ran a mission, all went great, got an Air Medal, but when I exited DCS and came back to Retribution (Alpha 2406) I could not "Accect Results", only choices were Manual and Abort. What did I miss?

>if you get a chance can you take a look at this? This has been reported a few times since my changes and then your subsequent fix

https://discord.com/channels/1015931619187621999/1015931619669983256/1474058613336178699

### Save game and other files

none

red-one1:
## Summary

This PR hardens the post-mission `state.json` pipeline end-to-end (Lua writer, trigger producers, Python parser, and UI polling) to prevent cases where mission results are not persisted or the UI never enables **Accept results**.

It builds on PR-660 mission-end write behavior and closes additional reliability gaps discovered from field reports.

Fixes #697

### 1) Mission-end write could fail once and never persist final state
**Issue**  
At mission end, direct state write attempts could fail (for example, transient Windows file lock contention while Python is reading), which could prevent the final `mission_ended=true` state from being saved.

**Change and how it works**  
- Mission-end write is now wrapped in a protected call (`pcall`) in `resources/plugins/base/dcs_retribution.lua`.
- If the write succeeds, `dirty_state` is cleared.
- If mission has ended and state is still dirty, scheduled retries happen every 1 second (instead of only normal cadence), so final state converges quickly.

---

### 2) Debrief polling race could miss the final mission-ended update
**Issue**  
The polling thread used to emit one update and exit; the UI then spawned a new thread. If DCS wrote the final update in that gap, the UI could stay on “Mission is being played” with only Manual/Abort actions.

**Change and how it works**  
- `PollDebriefingFileThread` now continues polling until `mission_ended=true`.
- The UI no longer recreates polling threads on every intermediate update in `QWaitingForMissionResultWindow`, removing the handoff race window.

---

### 3) Partial/in-flight `state.json` reads could break debrief parsing
**Issue**  
If Python read `state.json` during a write, missing keys could raise exceptions and kill/degrade result processing.

**Change and how it works**  
- `StateData.from_json` now uses safe defaults (`.get(..., [])` and fallback boolean) in `game/debriefing.py`.
- Polling now catches transient read/parse failures (`JSONDecodeError`, `OSError`, `ValueError`, `KeyError`) and retries.

---

### 4) Some trigger-generated events were never marked dirty
**Issue**  
Several mission triggers appended to state tables but did not set `dirty_state=true`, so those updates could be skipped by scheduled writes.

**Change and how it works**  
Added `dirty_state = true` to trigger scripts that mutate state arrays:
- `game/missiongenerator/tgogenerator.py`
- `game/missiongenerator/triggergenerator.py`
- `game/pretense/pretensetriggergenerator.py`

This guarantees scheduler visibility for those mutations.

---

### 5) Lua writer robustness at open/write boundary
**Issue**  
`write_state()` needed stronger failure handling around file open and JSON availability.

**Change and how it works**  
- `json` availability is now checked before opening the file handle.
- `io.open` now checks for failure explicitly and raises with path/error details.
- This avoids attempting writes when prerequisites are unavailable and improves diagnosability of write failures.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -30,7 +30,7 @@
 * **[Modding]** Added VSN F-35A/B/C mod support
 * **[Modding]** Add F-111C Aardvark by Warpig Productions (v2.260208)
 * **[Data]** Added ability to restrict weapons usage for a faction to a different year from the nominal weapon introduction year. Updated faction data to restrict more advanced missiles from Soviet client states during the cold war. Updated Egypt 2000 faction to restrict AIM-120 usage.
-* **[Modding]** Update OH-6A mod to v1.6
+* **[Modding]** Update OH-6A mod to v1.7
 
 ## Fixes
 * **[Flight Plans]** Fixed bugs wrt planning escort flights
--- a/game/data/weapons.py
+++ b/game/data/weapons.py
@@ -104,8 +104,11 @@ class Weapon:
 
     def available_on(self, date: datetime.date, faction: Faction) -> bool:
         introduction_year = self.weapon_group.introduction_year
-        if self.weapon_group.name in faction.weapons_introduction_year_overrides:
-            introduction_year = faction.weapons_introduction_year_overrides[
+        faction_introduction_year_overrides = getattr(
+            faction, "weapons_introduction_year_overrides", {}
+        )
+        if self.weapon_group.name in faction_introduction_year_overrides:
+            introduction_year = faction_introduction_year_overrides[
                 self.weapon_group.name
             ]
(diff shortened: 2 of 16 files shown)


Raffson (approved):
Lgtm :+1:

PR #699, opened 2026-02-24
Captured airfields will get TACANs

red-one1:
Added an option - When OWNFOR captures an airfield, if it doesn't normally have a TACAN, this will give it one.

- Implement PortableTacanGenerator to place TACAN beacons at airfields without built-in TACANs.
- Introduce a setting to enable automatic placement of portable TACANs at blue-captured airfields.
- Update control point logic to check for dynamic runway overrides.
- Add unit tests for PortableTacanGenerator, covering callsign derivation and generation conditions.

### DIFF
--- a/game/missiongenerator/tgogenerator.py
+++ b/game/missiongenerator/tgogenerator.py
@@ -53,7 +53,7 @@ from dcs.triggers import (
 from dcs.unit import Unit, InvisibleFARP, BaseFARP, SingleHeliPad, FARP
 from dcs.unitgroup import MovingGroup, ShipGroup, StaticGroup, VehicleGroup
 from dcs.unittype import ShipType, VehicleType
-from dcs.vehicles import vehicle_map, Unarmed
+from dcs.vehicles import vehicle_map, Unarmed, Fortification as VehicleFortification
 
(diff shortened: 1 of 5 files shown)


game/missiongenerator/tgogenerator.py
```diff
@@ -1367,6 +1376,120 @@ def generate(self) -> None:
...
+            logging.warning(
+                "No TACAN channels available for portable beacon at %s",
+                self.airfield.name,
+            )
+            return
+
+        callsign = self._derive_callsign(self.airfield.name)
+
+        # Place the portable TACAN beacon near the airport reference point.
+        position = airport.position.point_from_heading(
```
Raffson, line 1428: Is the TCN-beacon a physical object in-game? If so, how do we guarantee it doesn't spawn on the runway? :sweat_smile:

Raffson: Looking more closely, offsetting it 50m perpendicular to the runway seems to make sense, and I guess the edge-case where you'd have runways that cross each other perpendicular won't occur in practice 🤔

Raffson (approved):

PR #731, opened 2026-04-16
Update SEAD Waypoint Tasks

Drexyl:
Updated SEAD Waypoint Tasks in line with working DEAD Waypoint Tasks.

### DIFF
--- a/game/missiongenerator/aircraft/waypoints/seadingress.py
+++ b/game/missiongenerator/aircraft/waypoints/seadingress.py
@@ -1,4 +1,5 @@
 import logging
+import math
 
 from dcs.point import MovingPoint
 from dcs.task import (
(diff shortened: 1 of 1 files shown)


game/missiongenerator/aircraft/waypoints/seadingress.py
```diff
@@ -54,22 +55,42 @@ def add_tasks(self, waypoint: MovingPoint) -> None:
...
-                engage_task = EngageGroup(miz_group.id)
-                engage_task.params["weaponType"] = DcsWeaponType.ARM.value
-                engage_task.params["expend"] = Expend.All.value
-                waypoint.tasks.append(engage_task)
+            attack_task = AttackGroup(
+                miz_group.id,
+                weapon_type=DcsWeaponType.ARM,
+                expend=Expend.All,
+                altitude=waypoint.alt,
+                group_attack=True,
```
Raffson, line 63: In combination with `Expend.All` this will likely result in all HARMs being fired at one target only. In many cases this would probably suffice but this will perform really bad when the target consists of TELARs 🤔

game/missiongenerator/aircraft/waypoints/seadingress.py
```diff
@@ -54,22 +55,42 @@ def add_tasks(self, waypoint: MovingPoint) -> None:
...
+            waypoint.tasks.append(attack_task)
 
-            # Use other Air-to-Surface Missiles at last
             attack_task = AttackGroup(
                 miz_group.id,
                 weapon_type=DcsWeaponType.ASM,
-                altitude=waypoint.alt,  # flight loses alt with AB restriction
+                expend=Expend.All,
+                altitude=waypoint.alt,
+                group_attack=True,
```
Raffson, line 72: Same problem here, except this will start firing ASMs at the second target of the group...

game/missiongenerator/aircraft/waypoints/seadingress.py
```diff
@@ -54,22 +55,42 @@ def add_tasks(self, waypoint: MovingPoint) -> None:
...
+
+            dir = target.position.heading_between_point(waypoint.position)
+
+            attack_task = AttackGroup(
+                miz_group.id,
+                weapon_type=DcsWeaponType.Unguided,
+                attack_limit=1,
+                expend=Expend.All,
+                direction=math.radians(dir),
+                altitude=waypoint.alt,
```
Raffson, line 93: Doesn't this make SEAD effectively become DEAD, except for the fact that everything gets dumped on 1 target? 🤔
I can understand the `Expend.All` setting here in order to have the AI bug out ASAP, but enabling group-attack for guided bombs is still questionable imo 🤔

Starfire13: Well, considering we currently have to use DEAD as a replacement for SEAD until SEAD is fixed, that's not really a bad thing is it.

The whole point of this PR is so that we get back a functional SEAD.

Against TELARS, SEAD Sweep is very effective and I use that instead of SEAD. But we need SEAD to work against LORADS like they used to, where they have to be killed via saturation attacks on the radar.

Raffson: Perhaps a stupid idea, but what if we set the Visible option in the search task if we notice that the target is LORAD? 🤔

Starfire13: What happens if we do that?

Raffson: AI will fire off their HARMs in PB mode basically

Starfire13: That is fine with me. As long as there is a way for AI to perform saturation attacks rather than just firing a single HARM at a LORAD while it laughs.

Raffson: I should test it for good measure, but as I recall it ought to work. Although one known issue is that the HARMs could get guided to a different target, so I think it should only apply to LORAD since the changes of a HARM locking onto a different target is rather small, and when it does it shouldn't be a big issue because I'd assume the target will be in the same general area 🤔

Starfire13: Why not just revert to the way it used to work then as it was working perfectly? That way we have SEAD for LORADS (and potentially BUKs) and SEAD Sweep for everything else.

Raffson (approved):
Time doesn't seem to be on my side lately so let's merge this as a temporary fix...

PR #735, opened 2026-05-08
Allow per-campaign setting of player laser-code preferences, unify laser code handling

geofffranks:
- Collapses `FlightMember.tgp_laser_code` and `FlightMember.weapon_laser_code` into a single `laser_code` field. The same value now drives the kneeboard, the in-cockpit TGP code, and per-pylon LGB laser-code settings, so they cannot disagree.
- Adds a campaign setting on the Mission Generator page — `Default laser code for Player flights` with two values: `Default (1688)` and `Allocate own (unique per flight)`. Default is `Allocate own`, matching prior shipping behaviour. Also gets applied automatically when a slot is toggled from AI to player.
- Replaces the dual "Assigned TGP laser code" + "Preset laser code for weapons" controls in the payload tab with a single `Laser code:` selector. JTAC codes remain available as a per-flight override.
- Forward-only `__setstate__` migration handles every legacy save shape (only-tgp, weapon == tgp, weapon = JTAC with orphan tgp, weapon = `None` with orphan tgp). Orphan reservations are returned to the registry so the pool does not leak.

## Test Plan

- [x] Unit tests for the ownership-aware helpers, `FlightMembers` release semantics, and the campaign-default flight-creation path. New: `tests/ato/test_flightmember.py`, `tests/ato/test_flightmembers.py`, `tests/commander/test_packagebuilder.py`.
- [x] Pickle round-trip migration tests covering each legacy save shape and idempotency on already-migrated saves: `tests/ato/test_flightmember_migration.py`.
- [x] Full suite: `pytest tests` — 101 passed, 2 skipped.
- [x] `black --check` clean across all changed files.
- [x] Load an existing campaign written by a previous release; verify all player flights still have valid laser codes after migration.
- [x] Create a new player flight in the auto-planner; verify `member.laser_code` is set and the F-16C `LaserCode1/10/100` properties end up in the generated `.miz`'s `AddPropAircraft` block.
- [x] Toggle a slot from AI to player on an existing flight; verify no `AttributeError` and that the new player gets a code applied per the campaign setting.

## Notes for reviewers

- The `_migrate_legacy_laser_code_fields` helper reads the original `state` dict (not `self.__dict__`) so the constructor's seeding does not confuse new-format saves with legacy ones.
- Cockpit TGP pre-set only works for aircraft whose DCS module exposes the relevant `AddPropAircraft` keys — currently F-16C and F-15E in this repo. A-10C / A-10C II / F/A-18C have no such mission-file knob; the pilot still has to dial the code in cockpit. Out of scope for this PR.


Druss99:
We don't actually want to collapse these settings as some users may want to plan a flight where one friend lases for another

geofffranks:
> We don't actually want to collapse these settings as some users may want to plan a flight where one friend lases for another

I'm pretty sure that's still possible with these changes, in multiple ways:
- changing the lasing plane's TGP to match the bombing's plane's laser code set on the bombs
- by having the bombing plane use LSS on the lasing plane's code, then re-acquiring its own lock and lasing with its own code
- by setting the flight to use the default 1688 instead of player-specific codes
- by setting the default for all flights to use default-1688 instead of player specific codes

Also, on the F-16/A-10 at least, I haven't seen the TGPs ever start up reflecting what the player's assigned laser code is, it's always defaulting to 1688 regardless of what retribution sets. Maybe this is different on other planes, but to me it didn't seem like "Assigned TGP laser code" was actually honored anywhere in DCS.

Druss99:
> > We don't actually want to collapse these settings as some users may want to plan a flight where one friend lases for another
> 
> 
> 
> I'm pretty sure that's still possible with these changes, in multiple ways:
> 
> - changing the lasing plane's TGP to match the bombing's plane's laser code set on the bombs
> 
> - by having the bombing plane use LSS on the lasing plane's code, then re-acquiring its own lock and lasing with its own code
> 
> - by setting the flight to use the default 1688 instead of player-specific codes
> 
> - by setting the default for all flights to use default-1688 instead of player specific codes
> 
> 
> 
> Also, on the F-16/A-10 at least, I haven't seen the TGPs ever start up reflecting what the player's assigned laser code is, it's always defaulting to 1688 regardless of what retribution sets. Maybe this is different on other planes, but to me it didn't seem like "Assigned TGP laser code" was actually honored anywhere in DCS. 

I still don't see how collapsing the two settings in to one is beneficial. We can control both settings and make them the same by default so I see no reason to collapse them and remove that customization from users

geofffranks:
> I still don't see how collapsing the two settings in to one is beneficial. We can control both settings and make them the same by default so I see no reason to collapse them and remove that customization from users

👍 Ok, i pulled that part out of the PR. should now be the two configs again.

Druss99 (approved):

PR #740, opened 2026-05-18
Parking space information improvements

juanjux:
## Changes

- The (very annoying) "insufficient parking" dialog is now non modal and groups all this warnings into a single dialog that names the missing parking type (I was very confused about this at the start because I didn't notice parking was split into types!) and make the squadron text clickable so it will open the squadron dialog to fix the issue.
- Base intel summary in the airbase dialog shows aircraft occupied / transferring / free and per-parking type breakdown (shared / fixed wing exclusive / rotary wing exclusive / ground pawns) fixing the misleading "0 fixed wing only parking" at bases whose slots are all shared. For fixed wing aircraft, it also divides the information into small and big slots.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,7 @@
 * **[Options]** New option to spawn TACAN beacons at captured airfields
 * **[AirWing]** Track per-squadron campaign aircraft stats (initial/destroyed/purchased, save-compatible) and expose pilot experience level and living/dead pilot views for the UI
 * **[AirWing]** Squadron list shows living-pilot, aircraft and unassigned counts; squadron dialog shows each pilot's experience level, lists killed-in-action pilots separately and hides the redundant "Active" status
+* **[AirWing]** Squadron dialog shows the aircraft type, an aircraft inventory (initial/current/destroyed/purchased), and buy/sell aircraft controls with price, on-order count and available parking slots
 
 ## Fixes
 * **[Performance]** Improved robustness w.r.t. state.json handling to avoid corruption and thus save loss.
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@
 * **[Modding]** Add support for Su-35S mod (v2.0.27b)
 * **[Plugins]** Update EW Script to version 2.1
 * **[Options]** New option to spawn TACAN beacons at captured airfields
+* **[AirWing]** Track per-squadron campaign aircraft stats (initial/destroyed/purchased, save-compatible) and expose pilot experience level and living/dead pilot views for the UI
 
 ## Fixes
 * **[Performance]** Improved robustness w.r.t. state.json handling to avoid corruption and thus save loss.
--- a/changelog.md
+++ b/changelog.md
@@ -12,6 +12,7 @@
 * **[Plugins]** Update EW Script to version 2.1
 * **[Options]** New option to spawn TACAN beacons at captured airfields
 * **[UX]** Show an "End of Mission Detected, processing Mission Data" busy dialog while turn results are processed, so the wait is not mistaken for a missed detection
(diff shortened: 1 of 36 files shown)


Druss99:
I'll try to review this one this week but the parking situation is quite complex due to there being a disconnect between how retribution calculates parking slots and how pydcs calculates them. You can check one of my previous PRs on this to see what I mean but TLDR; the parking calc on the pydcs side is very expensive so we try to only do it on mission generation

juanjux:
> I'll try to review this one this week but the parking situation is quite complex due to there being a disconnect between how retribution calculates parking slots and how pydcs calculates them. You can check one of my previous PRs on this to see what I mean but TLDR; the parking calc on the pydcs side is very expensive so we try to only do it on mission generation

As anecdotic experience, this has been working well for me on my fork, however I didn't knew about the pydcs problem. I am outside home (and my Windows gaming computer where I test retribution) this week on a work trip but will re-check this in the weekend. Will try to make this PR independent of the other air wing ones so it doesnt block 5 and 6.

juanjux:
@Druss99 since this PR will be on hold until I can review the pydcs problem (moved to draft), I have rebased https://github.com/dcs-retribution/dcs-retribution/pull/741 and https://github.com/dcs-retribution/dcs-retribution/pull/742 to `dev` without branching from this PR so they can be reviewed already (#742 better after merging #741 since it branches from it because it uses some stuff from #741).

juanjux:
> I'll try to review this one this week but the parking situation is quite complex due to there being a disconnect between how retribution calculates parking slots and how pydcs calculates them. You can check one of my previous PRs on this to see what I mean but TLDR; the parking calc on the pydcs side is very expensive so we try to only do it on mission generation

Thanks, that context helped. I dug into it — here's where the expensive pydcs slot calc actually runs in this PR:

The per-category breakdown only does the deepcopy(dcs_airport) + free_parking_slot() placement sim when the base menu dialog is opened (and on the cheat/repair buttons), via `update_intel_summary`, it only runs for the single base you opened. That's the same cost profile as the existing transfer-destination dropdown (`SquadronDialog.calculate_parking_slots`), which already does the same deepcopy + free_parking_slot on demand.

So the heavy calc stays bounded to user-initiated actions. If you'd still rather not pay it on every open, I can cache the result per base and invalidate on transfers/turn change. Does the on-open approach work for you, or would you prefer the cached version?

Druss99:
FYI I'm about to be out of town for a week but could you do some testing to see load times between the old system and your PR? If you can log the duration that would be nice since depending on your PC you might not notice a difference but those with lower specs could

juanjux:
> FYI I'm about to be out of town for a week but could you do some testing to see load times between the old system and your PR? If you can log the duration that would be nice since depending on your PC you might not notice a difference but those with lower specs could

Yes, I have in my TO-DO adding some internal profiling to try to optimize some parts that are slow-ish like the Air Wing dialog and others, so I can include this A-B comparison.

juanjux:
@Druss99 

Benchmarked it on Shiraz (worst case: 119 slots, ~116 aircraft transferring in), opening the base menu 5×. Both builds identical except the parking calc:

- **Old path:** ~0.1 ms
- **PR (deepcopy + placement sim):** ~30 ms median, 93 ms on the first (cold) open

So ~30 ms added per base-menu open here. Not noticeable on my machine, but on lower-spec hardware that could be 2–4× and start to show on big bases. Cost is dominated by the `deepcopy` of the whole airport on every open — happy to cache it / avoid the copy if you think it's worth it (personally I don't think it is).

juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/26

PR #741, opened 2026-05-18
Air Wing improvements: transfer improvements (PR 1/2)

juanjux:
## Description

Basically fixing some paper cuts on the UI of of squadron transfers.

## Changes

- Adds a "Transfer ordered to X" text on the Air Wing list window and the Air Field Command panel, if the unit is being transfered to another place.
- Airfield Command also shows the units being transferred into the current base.
- Fix: the transfer-destination parking figure now accounts for transfers already ordered into that base (it previously ignored them).

### DIFF
--- a/qt_ui/windows/SquadronDialog.py
+++ b/qt_ui/windows/SquadronDialog.py
@@ -188,7 +188,13 @@ class SquadronDestinationComboBox(QComboBox):
             )
             free_ground_spawns = cp.total_aircraft_parking(parking_type)
 
-            for s in cp.squadrons:
+            # Squadrons that will actually occupy this base next turn:
+            # those staying here (not relocating away) plus any relocating in.
+            occupants = [s for s in cp.squadrons if s.destination is None]
+            occupants += [
+                s for s in cp.coalition.air_wing.iter_squadrons() if s.destination == cp
+            ]
+            for s in occupants:
                 for count in range(s.owned_aircraft):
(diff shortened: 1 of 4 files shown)


juanjux:
Note: rebased to `dev` now that the previous ones have been merged, should be review-able now.

qt_ui/windows/SquadronDialog.py
```diff
@@ -188,7 +188,13 @@ def calculate_parking_slots(
             )
             free_ground_spawns = cp.total_aircraft_parking(parking_type)
 
-            for s in cp.squadrons:
+            # Squadrons that will actually occupy this base next turn:
+            # those staying here (not relocating away) plus any relocating in.
+            occupants = [s for s in cp.squadrons if s.destination is None]
```
Druss99: This will break if a squadron tries to transfer to a new CP and that CP is captured by the opposing side on that turn. In that scenario the squadron has to be able to return to the previous CP

Druss99 (changes requested):

qt_ui/windows/SquadronDialog.py
Druss99: IE any outgoing transfers must count against the available parking

juanjux: Good catch! This should be handled by the updated PR, PTAL.

Druss99 (approved):

PR #744, opened 2026-05-20
Fix Take Off hang when fast-forward stop condition is unreachable

juanjux:
## Summary

Clicking "Take off" hangs the app forever when the configured "Fast forward until..." state is one the player flight will skip — e.g. "Startup time" with a runway start.

## Fix

Detect the mismatch before fast-forwarding and ask the user, per flight: change the flight's start type, or halt fast-forward at the flight's actual start instead. Both choices are mission-only.

A hard tick ceiling on the fast-forward loop catches any other unreachable case as a back-stop.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,7 +11,7 @@
 * **[Options]** New option to spawn TACAN beacons at captured airfields
 
 ## Fixes
-* **[Fast-forward]** Detect when the configured "Fast forward until" stop condition is unreachable for a player flight's start type (e.g. "Player startup time" with a runway or air start, which skip the Startup state) and prompt before launching the mission: per mismatched flight, the user can either downgrade the flight's start type to match the setting, or halt fast-forward at the flight's actual spawn instead, for this mission only. A hard tick ceiling also caps any remaining unreachable case as a back-stop, so the app can no longer hang on Take Off.
+* **[Fast-forward]** Fix Take Off hang when the configured "Fast forward until" state is one the player flight will skip (e.g. "Player startup time" with a runway start); the user is now prompted to resolve the mismatch before launch.
 * **[Performance]** Improved robustness w.r.t. state.json handling to avoid corruption and thus save loss.
 * **[Flight Plans]** Stabilized waypoint solver debug GeoJSON coordinate precision to avoid platform-specific floating point drift in debug output.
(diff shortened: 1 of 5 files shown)


Druss99:
I think this might work better (which I cherry picked from liberation) but I'll leave it to Stoner to decide: https://github.com/dcs-retribution/dcs-retribution/pull/752

juanjux:
> I think this might work better (which I cherry picked from liberation) but I'll leave it to Stoner to decide: #752

Is not the same if I am understanding your PR right, it's for adding an option to fast forward to the IP waypoint, while this one is for asking the user the start point when the configured one and the mission one for the player diverge.

Druss99:
> > I think this might work better (which I cherry picked from liberation) but I'll leave it to Stoner to decide: #752
> 
> 
> 
> Is not the same if I am understanding your PR right, it's for adding an option to fast forward to the IP waypoint, while this one is for asking the user the start point when the configured one and the mission one for the player diverge.

It adds that as an option but also checks to see if the player is past the FF point already as would be the case if you have player flight set to takeoff from ramp but you are fast forwarding to cold start

juanjux:
> > > I think this might work better (which I cherry picked from liberation) but I'll leave it to Stoner to decide: #752
> > 
> > 
> > Is not the same if I am understanding your PR right, it's for adding an option to fast forward to the IP waypoint, while this one is for asking the user the start point when the configured one and the mission one for the player diverge.
> 
> It adds that as an option but also checks to see if the player is past the FF point already as would be the case if you have player flight set to takeoff from ramp but you are fast forwarding to cold start

Ah, then it certainly would also fix the main problem, the infinite wait/loop issue when the configured start of before the mission start. I personally prefer the program to ask me in case in this case so both features + the single duplicated fix could be merged, if Stoner likes both, we could start by merging yours with the fix and the IP fast start option, and then I could add the option to show the dialog and the dialog itself on a rebased form of this PR later.

juanjux:
Moving to draft until #752 is merged (then I will rebase this one to only add the start time chooser dialog when there is a difference between the options' configured one and the player mission one).

juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/24

PR #745, opened 2026-05-21
Fix Anti-Ship flights not engaging carrier groups

juanjux:
## Summary

Anti-Ship flights tasked against a carrier group never fired — the AI flew to the ingress point and turned back without launching a single missile.

## Cause

The flight plan routes the package to the carrier (`AntiShipFlightPlan` uses `NavalControlPoint.find_main_tgo()`), but the waypoint task builder resolved its target independently from `control_point.ground_objects[0]`. A naval control point can own several ground objects, and the first one isn't necessarily the carrier — it can be a different naval group that is already sunk and therefore never spawned into the mission. The resulting `AttackGroup` task referenced a group that doesn't exist in the `.miz`, so it was silently dropped and the flight had no target.

## Fix

Resolve the target the same way the flight plan does, via `find_main_tgo()`, so the `AttackGroup` task points at the carrier group that's actually generated. Also emit a warning when an Anti-Ship flight ends up with no attack task at all, so a future mismatch surfaces in the log instead of failing silently.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,7 @@
 * **[Options]** New option to spawn TACAN beacons at captured airfields
 
 ## Fixes
+* **[Mission Generation]** Anti-Ship flights now attack the carrier group the flight plan routes to instead of the control point's first ground object, so strikes against carrier groups no longer leave the AI without a target (it would fly to the ingress point and turn back without engaging).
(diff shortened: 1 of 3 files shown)


juanjux:
@Druss99 I see you enabled auto merge, however this still needs a review to be mergeable.

Druss99 (approved):

PR #746, opened 2026-05-21
Fix Anti-Ship flights not disengaging after launching missiles

juanjux:
## Summary

Anti-Ship flights kept flying toward the enemy fleet after launching their missiles instead of returning to base. It's most obvious with aircraft that carry only stand-off weapons (e.g. the S-3B Viking with two Harpoons): they fire, then keep boring in on the fleet and get shot down.

## Cause

The `AttackGroup` task generated at the ingress point has no completion condition, so the AI never considers the task finished and keeps re-engaging the group.

## Fix (suggested by @Druss99)

Set `OptRTBOnOutOfAmmo(Values.Antiship)` on the flight in `configure_anti_ship`, so the AI returns to base as soon as its anti-ship missiles are spent. This reuses the existing `configure_behavior(rtb_winchester=...)` plumbing that CAP and Fighter Sweep already use for AAM/Cannon.

Triggering on ammo state is more reliable than capping the number of attack runs (an earlier version of this PR used `attack_limit=1`, but in testing the AI sometimes ignored the cap and kept re-engaging — thanks to @Druss99 for suggesting the RTB-options approach).

### DIFF
--- a/game/missiongenerator/aircraft/waypoints/seadingress.py
+++ b/game/missiongenerator/aircraft/waypoints/seadingress.py
@@ -1,4 +1,5 @@
 import logging
+import math
 
 from dcs.point import MovingPoint
 from dcs.task import (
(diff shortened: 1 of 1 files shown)


Druss99:
Can we fix this in a better way by setting their RTB options?
(Sorry not at my PC to test or check their current options when they are generated)

juanjux:
> Can we fix this in a better way by setting their RTB options? (Sorry not at my PC to test or check their current options when they are generated)

Yes. 

I actually was about to close this PR because after some more testing it's still happening, just less, but your comment inspired my to try the RTB options instead. I have updated this same PR to do that and I am testing this right now but it seem to work a lot better in the 3 missions I have done so far.

juanjux:
Update: with your suggested change, plus another small one, in a mission I just played out of 10 flights in an anti ship strike package only one did the idiotic thing, which is much better than before where 30-50% would do it.

game/missiongenerator/aircraft/waypoints/antishipingress.py
```diff
@@ -29,15 +29,18 @@ def add_tasks(self, waypoint: MovingPoint) -> None:
...
+        # Harpoons) and gets shot down before it has fired its standoff
+        # weapons. Standoff-friendly categories only.
         self.add_attack_group_tasks_for_ordnance(
             waypoint, group_names, WeaponType.Antiship
         )
         self.add_attack_group_tasks_for_ordnance(
             waypoint, group_names, WeaponType.Guided
         )
-        self.add_attack_group_tasks_for_ordnance(
-            waypoint, group_names, WeaponType.Unguided
```
Druss99, line 39: This will break loadouts with Mk82s

game/missiongenerator/aircraft/aircraftbehavior.py
```diff
@@ -291,6 +291,11 @@ def configure_anti_ship(self, group: FlyingGroup[Any], flight: Flight) -> None:
             group,
             react_on_threat=OptReactOnThreat.Values.EvadeFire,
             roe=OptROE.Values.OpenFire,
+            # RTB once the anti-ship missiles are spent. Without this the
+            # AttackGroup task on the ingress waypoint has no completion
+            # condition, and aircraft carrying only stand-off weapons (e.g. the
+            # S-3B) keep flying in after launching their missiles.
+            rtb_winchester=OptRTBOnOutOfAmmo.Values.Antiship,
```
Druss99: Unsure if this will work for flights with Mk82s as in they might RTB immediately but unsure

Druss99 (changes requested):

game/missiongenerator/aircraft/waypoints/antishipingress.py
juanjux: Yeah. I was thinking that those kamikaze flights could be targeting the ships with the gun after expending the standoff weapons, but of course this would also affect builds with unguided bombs 🤦 .

I have removed it and will push it soon.

game/missiongenerator/aircraft/aircraftbehavior.py
juanjux: Yup. Removed, as said above will think of another approach.

PR #747, opened 2026-05-21
Fix two-column delegate crash in sizeHint under PySide6 6.4.x

juanjux:
## Summary

A crash dialog — `'PySide6.QtWidgets.QWidgetItem' object has no attribute 'width'` — could pop up when a list backed by `TwoColumnRowDelegate` (Air Wing / ATO panels, etc.) relayouts, for example while the post-mission debriefing window is open.

## Cause

`TwoColumnRowDelegate.sizeHint()` reads `option.decorationSize` and calls `.width()`/`.height()` on it. Under PySide6 6.4.x the option handed to `sizeHint()` during some relayouts carries a `decorationSize` that isn't a `QSize` (it can be a `QWidgetItem`), so the attribute access raises.

## Fix

Guard `icon_size()`: if `decorationSize` isn't a `QSize`, treat it as "no icon" (`QSize(0, 0)`) instead of crashing. Behaviour is unchanged in the normal case.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,7 @@
 * **[Options]** New option to spawn TACAN beacons at captured airfields
 
 ## Fixes
+* **[UI]** Avoid a crash dialog ("'QWidgetItem' object has no attribute 'width'") when a list using the two-column row delegate relayouts with a malformed style option under PySide6 6.4.x.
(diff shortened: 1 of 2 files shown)


Druss99 (approved):

juanjux:
@Druss99 solved conflicts, this should be mergeable now.

Druss99 (approved):

PR #748, opened 2026-05-21
Show carrier/LHA ship groups on the map like other naval groups

juanjux:
## Summary

On the map, a carrier/LHA control point only showed its name. Its escort ships — a separate ship group attached to the control point — were invisible: no per-ship tooltip, and (more dangerously) **no air-defense threat rings**, so a battered carrier group whose carrier had sunk but whose SAM escorts were still alive looked harmless. The status marker also turned red (destroyed) the moment the carrier itself sank, even with escorts still afloat and present in the generated mission.

## Cause

Regular naval groups are `TheaterGroundObject`s and are serialized to the map as TGOs, with their unit list and threat/detection ranges. A carrier's ship group is an `is_control_point` ground object, which is intentionally excluded from the TGO serialization (to avoid duplicating the control-point marker), so none of that information reached the map. Separately, `NavalControlPoint.status` reported `Destroyed` as soon as the carrier ship (the "runway") was gone, ignoring surviving escorts.

## Fix

- **Backend:** expose the carrier/LHA control point's surviving units and its escorts' threat/detection ranges on `ControlPointJs`. Make `NavalControlPoint.status` reflect unit survival — Functional / Damaged / Destroyed — instead of keying off the carrier ship alone. This is purely cosmetic; launch capability is still governed by `runway_is_operational()`.
- **Frontend:** the control-point tooltip now lists the ship group (carrier + escorts, with `[DEAD]` markers for losses), and the air-defense range layer draws the surviving escorts' threat/detection rings on the control point.

## Note

`client/src/api/_liberationApi.ts` is generated (`npm run regenerate-api`) but was hand-edited here since codegen needs a running server; it can be regenerated from the updated `ControlPointJs` schema.


juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/23

PR #749, opened 2026-05-21
Fix duplicate SEAD Escort and missing SEAD task for carrier targets

juanjux:
## Summary

When creating a flight against an enemy carrier/LHA, the task list showed SEAD Escort twice and offered no plain SEAD, unlike a regular naval group.

## Cause

`NavalControlPoint.mission_types()` yields `SEAD_ESCORT` for enemy targets, but the base `MissionTarget.mission_types()` already yields `SEAD_ESCORT` — so it appeared twice. It also never offered plain `SEAD`, even though a carrier group's escorts are SAM platforms (a regular naval group does offer SEAD).

## Fix

Yield `SEAD` instead of `SEAD_ESCORT` for enemy carrier/LHA targets, matching regular naval groups: SEAD becomes available and the duplicate disappears (`SEAD_ESCORT` still comes once from the base mission types).

Adds a test asserting that no mission target yields a duplicate `FlightType` from `mission_types()`, for either coalition (auto-discovers every concrete `MissionTarget` subclass — control points, ground objects, transfers, front lines — so new types are covered too).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,7 @@
 * **[Options]** New option to spawn TACAN beacons at captured airfields
 
 ## Fixes
+* **[Mission Planning]** Carrier/LHA targets now offer SEAD in the flight-task list and no longer list SEAD Escort twice (their escorts are SAM platforms, so they can be suppressed directly like any other naval group).
 * **[Performance]** Improved robustness w.r.t. state.json handling to avoid corruption and thus save loss.
 * **[Flight Plans]** Stabilized waypoint solver debug GeoJSON coordinate precision to avoid platform-specific floating point drift in debug output.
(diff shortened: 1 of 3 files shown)


Druss99:
This looks okay to me but my fear is that if the escorts are sunk and only the carrier is left then a SEAD mission could be planned and fail since the carrier has no radar. I'll leave it to Stoner to decide if we should guard against that

Starfire13:
At the moment, an antiship flight against a carrier will expend all their ordnance on the carrier. They only attack escorts if the carrier is already dead.

juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/22

PR #751, opened 2026-05-21
Make aircraft payload handling robust (skip unparseable files; write loadouts atomically)

juanjux:
Two related robustness fixes for how aircraft payload `.lua` files are handled:

- **Reading:** an unparseable payload file (e.g. a hand-written third-party mod file like the CJS Super Hornet's FA-18E / EA-18G) no longer leaves the airframe with **no selectable payloads** — only the bad file is skipped, the rest (including your own Mission-Editor-saved loadouts) still load. If none can be read, the aircraft falls back to an empty loadout instead of aborting turn generation.
- **Writing:** loadouts are saved atomically (temp file + `os.replace`), so a reader never sees a half-written file.

Supersedes #796 (folded in here).


juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/21

PR #755, opened 2026-05-25
Migrate legacy fast-forward settings on load

geofffranks:
Resolves an issue when migrating campaigns from 1.5.0 to a commit after #684. That PR updated FastForwardStopCondition and CombatResolutionMethod enums but were missing migrations for existing saves. As a result loading from a save in the old format and trying to configure the campaign to update to the new settings would result in an error and the inability to change any settings. 

Complements #744, which solves an issue during take-off around the same circumstances. Both appear to be necessary.

### DIFF
--- a/game/settings/settings.py
+++ b/game/settings/settings.py
@@ -1465,14 +1465,110 @@ class Settings:
     def deserialize_state_dict(state: dict[str, Any]) -> dict[str, Any]:
         # restore Enum & timedelta types
         s = Settings()
-        for key, value in state.items():
-            if isinstance(s.__dict__.get(key), timedelta) and isinstance(value, int):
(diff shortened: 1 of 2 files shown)


Druss99:
I was unable to reproduce this issue, do you have a save file I can test with?

geofffranks:
not since I ran the migration here and it fixed my save.  I'll try to regenerate one and validate off the latest dev commit, then post here if it's still an issue.

geofffranks:
Ok, here's a reproduction save -  load, and click on the campaign settings icon, it should throw the error immediately.
[pr-755-reproduction.retribution.zip](https://github.com/user-attachments/files/28691389/pr-755-reproduction.retribution.zip)

Druss99 (approved):

PR #758, opened 2026-05-29
Improve auto-asigned TACAN codes' visibility

juanjux:
#699 from @red-one1 added portable TACANs to captured based, which is great and works beautifully. I found, however, that is not very obvious to find the TACAN code for a base; it was only added to the briefing and only if the airport was configured as a divert one. There was no way of knowing the code from the F10 map or the UI (if it was auto assigned).

This PR tries to improve the base's TACAN code visibility in these ways:

- Map tooltip: hovering a friendly base shows its TACAN and ATC frequency.
- Base / carrier dialog: the current TACAN channel and callsign are shown, with AUTO / AUTO (94X) labels so it's clear when it was auto-allocated vs user-set.
- Briefing: new Owned Airbases section listing every friendly airfield with its TACAN and ATC frequency.
- F10 map: the portable TACAN ground unit is named with its channel and callsign so you can read it straight from DCS.
- Carrier TACANs stay stable across turns instead of being re-rolled every mission generation.

Also, semi unrelated, I added that if you pick a channel that's already used by another base, carrier or flight, a confirmation dialog warns you before double-booking.


juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/20

PR #752, opened 2026-05-22
Make fast forward more robust and allow skipping combat

Druss99:
Cherry picked from liberation

PR #760, opened 2026-05-31
Fix player ground-start flights spawning in the air

juanjux:
## Summary

A player flight configured with a ground start (Cold/Warm/Runway) could spawn **in the air** when its computed `startup_time()` fell *before* the mission start time (e.g. when a long player startup estimate pushes startup earlier than the stored TOT allows).

In `FlightState.reinitialize()`, when `startup_time() <= now` the flight was pre-activated via `_set_active_flight_state(...)`, which skips the `StartUp`/`Taxi`/`Takeoff` state at tick 0. The `PLAYER_STARTUP` fast-forward stop condition only halts flights that are *currently in* `StartUp`, so it never caught these pre-activated flights and the simulation ran them straight through to `Navigating` — leaving the player airborne at mission start despite a cold/warm/runway start being selected.

This change keeps such flights grounded: a client flight with a ground start whose `startup_time()` lands at or before mission start now waits via `WaitingForStart(now)` and starts on the ground at mission start. AI flights and in-flight starts keep their previous behaviour.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,7 @@
 * **[Options]** New option to spawn TACAN beacons at captured airfields
 
 ## Fixes
+* **[Flight Plans]** Player flights with a ground start (Cold/Warm/Runway) no longer spawn in the air when their computed startup time falls before mission start (e.g. due to a long player startup estimate); they now wait and start on the ground at mission start.
 * **[Performance]** Improved robustness w.r.t. state.json handling to avoid corruption and thus save loss.
 * **[Flight Plans]** Stabilized waypoint solver debug GeoJSON coordinate precision to avoid platform-specific floating point drift in debug output.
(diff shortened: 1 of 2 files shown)


Druss99:
This is fixed by: https://github.com/dcs-retribution/dcs-retribution/pull/752

juanjux:
> This is fixed by: #752

Thanks, once #752 is merged I will rebase this removing the fix and leaving the dialog.

Druss99:
Also what is your discord handle?

juanjux:
> Also what is your discord handle?

@Druss99 gorthol7

juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/19

PR #766, opened 2026-06-06
DEAD/SEAD: one waypoint per target, with kneeboard STPT column

juanjux:
## Summary
Strike flights already build one `TARGET_POINT` waypoint per target unit, so every target the kneeboard lists with coordinates is also a selectable waypoint in the aircraft — trivial to designate with TOO. DEAD and SEAD flights listed the same per-unit targets (with coordinates) on their kneeboard page, but only generated a single target-area waypoint, so the player had to enter each target by hand.

This makes DEAD and SEAD flights against a ground objective generate one `TARGET_POINT` waypoint per target, exactly like Strike, and adds the matching **STPT** column to the SEAD/DEAD kneeboard target list.

## Behaviour / compatibility
- **AI tasking is unchanged**: the generated waypoints are `only_for_player`, so the AI still attacks via the `AttackGroup` tasks on the ingress point.
- Additive and backwards-compatible: existing flight plans without per-target waypoints simply show a blank STPT; non-ground SEAD targets are unaffected.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,7 +9,7 @@
 * **[Modding]** Add support for Su-35S mod (v2.0.27b)
 * **[Plugins]** Update EW Script to version 2.1
 * **[Options]** New option to spawn TACAN beacons at captured airfields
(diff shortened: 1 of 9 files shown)


Starfire13:
This should be for player flights only and is just to make slaving to targets easier, yeah? AI flights don't require target waypoints when attacking a unit group.

juanjux:
> This should be for player flights only and is just to make slaving to targets easier, yeah? AI flights don't require target waypoints when attacking a unit group.

That's correct, this doesn't touch AI tasking at all.

juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/18

PR #767, opened 2026-06-06
Relocate ground units that spawn on water to nearest land (#59)

### LINKED ISSUE #59: Some armor spawns under water
### Affected versions

Development build

### Test/Development build

https://github.com/dcs-retribution/dcs-retribution/actions/runs/3814230005

### Description

Several groups spawned under the Sea of Gallilee. There were some boats there initially. Once they were killed it was discovered several groups under the water. This should be easy to repeat by loading the .miz. 

### Save game and other files

Liberation Save, Track and Miz
[DCS.openbeta.zip](https://github.com/dcs-retribution/dcs-retribution/files/10328416/DCS.openbeta.zip)
[DCS.openbeta (2).zip](https://github.com/dcs-retribution/dcs-retribution/files/10328417/DCS.openbeta.2.zip)

Rename second file to "DCS.openbeta.z01" before extracting.

geofffranks:
### Problem
Ground units sometimes spawn on water — inland lakes/rivers or right at the
coastline — leaving them stranded or sunk and effectively dead on arrival.

### Fix
A small runtime Lua "nudge" in the mandatory **base** plugin
(`resources/plugins/base/water_relocate.lua`). At mission start it finds ground
units sitting on `WATER`/`SHALLOW_WATER` and respawns each affected group on the
nearest land (expanding-ring search, ≤ 2000 m), preserving the group's name,
units, route, tasks and skill.

- **Theater-agnostic** — keys off DCS's runtime `land.getSurfaceType`, not
  Retribution's landmap.
- **Loss-safe** — re-adds the group under the same name via
  `mist.getGroupData` + `mist.dynAdd`; `coalition.addGroup` with an existing
  name fires `S_EVENT_BIRTH` only, so none of the kill/loss events Retribution
  tracks are emitted.
- Registered last in `plugin.json` (after mist, which it requires).
- Units with no land within 2000 m are left in place with a warning.

### Testing
- **Automated (CI):** new `tests/missiongenerator/test_water_relocate_plugin.py`
  (registration, load-order-after-mist, DoScriptFile injection). All gates green
  — `black`, `mypy game`, `mypy tests`, `pytest`. A standalone Lua unit test of
  the real script under a mocked DCS/mist env also passes under both `lua` and
  `luajit`.
- **In-DCS acceptance (Caucasus):**
  - *Open water (Black Sea):* wet units relocate to shore; a far-offshore unit
    (>2 km from land) is correctly left in place + warned; on-land units
    untouched.
  - *Inland (Tbilisi Sea reservoir):* units spawned on the lake relocate to the
    shoreline; on-land units untouched.
  - *Loss-safety:* every relocated, un-engaged unit is alive at mission end —
    `dcs.log` shows the expected `relocating …` lines, and the debrief shows
    no spurious losses (no deaths, no graveyard entries).

Fixes #59.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -29,6 +29,7 @@
 * **[Flight Plans]** Stabilized waypoint solver debug GeoJSON coordinate precision to avoid platform-specific floating point drift in debug output.
 * **[Mission Generation]** Assign plane-specific laser codes to LGB weapons when building the mission
 * **[Engine]** Fixed a bug where squadrons could transfer to enemy owned control points
+* **[Mission Generation]** Relocate ground units that spawn on inland water (ponds, rivers, lakes) to the nearest land at mission start, fixing armor and SAM sites spawning underwater (#59).
 
 # Retribution v1.5.0
 
--- a/changelog.md
+++ b/changelog.md
@@ -30,6 +30,7 @@
 * **[Mission Generation]** Assign plane-specific laser codes to LGB weapons when building the mission
 * **[Engine]** Fixed a bug where squadrons could transfer to enemy owned control points
 * **[Mission Generation]** Relocate ground units that spawn on inland water (ponds, rivers, lakes) to the nearest land at mission start, fixing armor and SAM sites spawning underwater (#59).
+* **[Mission Generation]** Relocate ships that spawn on land (e.g. carrier escorts when the carrier hugs the shore) to the nearest deep water at mission start (#59).
 
 # Retribution v1.5.0
 
--- a/resources/plugins/base/land_relocate.lua
+++ b/resources/plugins/base/land_relocate.lua
@@ -51,6 +51,14 @@ local SEARCH_HEADINGS = 8        -- sample points per ring (every 45 degrees)
 local CLEAR_RADIUS = 1 * NM      -- destination must be open water this far out...
 local CLEAR_STEP   = 0.2 * NM    -- ...sampled every 0.2 nm along each spoke
 
+-- Relaxed fallback: when the strict open-water search finds nothing (e.g. a
+-- carrier's escorts in a narrow canal), a beached ship is moved to the nearest
+-- deep water. The gated relaxed pass keeps that destination at least
+-- MIN_ANCHOR_DIST from the nearest afloat ship so escorts don't stack on the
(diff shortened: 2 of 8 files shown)


juanjux:
Not sure if it could make sense to handle this on this PR or another one, but I have also seen the reverse problem - ships (escort ships from a carrier) spawning on land when the carrier is close to the shore.

geofffranks:
@juanjux I pushed up a commit that should also handle this for ships that spawn on land (slightly differently). It makes sure they have deep water around them in at least a 1 mile radius so they can be sure they're able to move properly if so tasked. it also widens the search radius to 7nm to account for how far they might spawn inland as part of a carrier group.

juanjux (approved):

juanjux:
@geofffranks nice! I have reviewed the PR and didn't find anything else to improve, but as a very late contributor I would wait for a review from @Druss99 or @Starfire13

Druss99:
Looks pretty good to me but have we tested the Suez Canal? I think there's a campaign with a carrier group in it and it is a tight squeeze. I'm also slightly worried about BAI and strike tasks against anything that gets relocated as I'm not sure how AI would handle that

geofffranks:
I have not tested it on Suez. I imagine it will probably have issues there. Will check it out. 

BAI and Strike tasks probably don't matter as much on the ground units, those are typically front line units being spawned in the wrong spots, and the BAI/Strike targets are all vetted by the campaign designer, right?

But for water I get the concern.  I'll see what I can test there  too.

Druss99:
That makes sense

geofffranks:
Tested the Suez.  In the campaign with ships in it, they start with only the LHA carrier in the canal which doesn't seem to get any escorts (i think?) - it's able to be in the canal on its own just fine. 

I moved the CVN-72 into the canal and its escorts spawn on the ground surrounding the CVN-72. When I tested the same scenario on the dev branch, they didn't show up at all. I'll see if I can get them to fall back to just finding the nearest water even if it doesn't match  the initial search parameters

Starfire13:
The Sinai campaign with the LHA in the canal is one of mine. Yep, it's fine there on its own, as you found. No escorts.

Druss99 (approved):

PR #771, opened 2026-06-07
Use Qt non-native dialogs to avoid QtWebEngine file-dialog deadlock

juanjux:
## Problem
Opening a **native Windows file dialog while the embedded web map (`QWebEngineView`) is alive** can deadlock the whole app and freeze it ("Not Responding"): the Windows platform plugin issues a synchronous `SendMessage` that never returns while the (nested, modal) dialog is created.

Reproduced by clicking **"Manually submit"** on the modal "Waiting for mission completion" window (`QWaitingForMissionResultWindow.submit_manually` → `QFileDialog.getOpenFileName`). A `py-spy --native` dump of the hung process showed the main thread blocked under `QDialog::exec` at `NtUserMessageCall` / `GetSystemMetrics` / `CallWindowProcW` in `qwindows.dll`, with the Qt event loop no longer pumping messages.

It's not limited to that one button — every native dialog opened over the map is at risk.

## Fix
`AA_DontUseNativeDialogs`: file / colour / font choosers use Qt's own (non-native) implementations instead of the deadlocking native path.

## Trade-off
File dialogs use the Qt-styled chooser instead of the native Windows one.


juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/17

PR #772, opened 2026-06-08
feat(sead): loiter-and-react SEAD with standoff orbit

geofffranks:
Currently SEAD flights fly out to SEAD Search, do a point in time check for targets to suppress, and engage, then RTB. If none of the radars are online yet because nothing has gotten within the threat zone, they immediately RTB without doing anything. This updates the behavior so they set up an orbit at SEAD Search, engaging threats as they come online. Once they're winchester, or their flight plan's on-station time is up, they RTB.
 
Loiter and reactive engagement:
- Emit a SEAD_LOITER waypoint at factor x threat-range standoff, capped at 95% of the ingress distance and floored (MIN_SEAD_STANDOFF_FACTOR) so a misconfigured 0 can never park the orbit on top of the SAM.
- SeadLoiterBuilder injects a bounded circular orbit via a shared add_stopping_orbit helper (also used by the hold-point builder); the stop time is floored at 1s so a 0 window or an already-elapsed TOT cannot kill the orbit at mission start.
- Drop the point-in-time AttackGroup tasks from the SEAD ingress; keep ECM on for survivability during the loiter.
- Add sead_loiter_standoff_factor (0.1-2, default 0.8) and sead_loiter_max_window_seconds (default 1200) doctrine settings.
- Draw a fixed 20 NM HARM-reach engagement zone for plain SEAD in the planner, centered on the loiter anchor, independent of the user-tunable SEAD-sweep engagement range.
- Breaks off the loiter when winchester, or when time is up based on calculating the flight plan of the package (similar to how escort is planned)

Tests cover the standoff geometry and factor floor, the loiter stop-time guard, the engagement-zone constant, and the new settings defaults.

### DIFF
--- a/game/ato/flightplans/sead.py
+++ b/game/ato/flightplans/sead.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from datetime import timedelta
+from datetime import datetime, timedelta
 from typing import Type
 
 from .formationattack import (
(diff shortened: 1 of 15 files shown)


Starfire13:
Out of ARM isn’t the right condition type because SEAD Search is often used for targeting non-radar SAMs. In fact, many aircraft capable of SEAD Search carry mixed loadouts. I often use SEAD Search flights with Mavericks to clean up SAM sites that have already been degraded by a SEAD flight, for example. It's a role they are very effective at.

geofffranks:
I think there's a little confusion here. This change is only for the SEAD flight. It has a specific waypoint called "SEAD Search" (Now being renamed to "SEAD Loiter"). The SEAD Sweep flight sounds like what you're talking about, with a SEAD Sweep waypoint. This flight type is unchanged by this PR.

Starfire13:
Oh sorry, yeah I thought this was referring to SEAD Sweep. My bad!

Druss99:
This seems overly complex to achieve the stated goal. I think you should start with a problem you experience and what would solve that problem in a simple way. When it comes to flight tasking we normally try to work backwards from the mission editor. IE, create the scenario in the mission editor, find out how to best task the AI, and then implement

geofffranks:
> This seems overly complex to achieve the stated goal. I think you should start with a problem you experience and what would solve that problem in a simple way. When it comes to flight tasking we normally try to work backwards from the mission editor. IE, create the scenario in the mission editor, find out how to best task the AI, and then implement

What part are you finding overly complex, the logic around when the SEAD flight should  RTB, or something else? Would you be more ok with pulling the friendly-flights-in-package leaving the threat zone out, but leaving int the time-based + ammo-based triggerd to RTB?

PR #776, opened 2026-06-09
Fix IndexError crash when flight exits combat at last waypoint

BradySox:
next_waypoint_state() was calling Navigating(new_index) even when new_index was the last waypoint in the plan, causing InFlight.__init__ to crash on waypoints[new_index + 1].

Triggered by SCRAMBLE flights whose BarCap plan exits combat at or past the final waypoint before a LANDING_POINT is reached.

Fixes both layers:
- next_waypoint_state(): return Completed when new_index+1 is out of bounds rather than constructing a Navigating that will immediately crash
- InFlight.__init__: replace the TODO comment with a clear ValueError so any future call site gets a useful message instead of IndexError

Pull requests should be made against the `dev` branch. Any backports
necessary will be handled by the development team.

Pull requests should be focused on one task. Multiple bug fixes should be
multiple PRs, or at the very least split over multiple (isolated) commits.
We cannot merge half a PR, and combined PRs are more
difficult to review. PRs that do not adhere to this may have their review
delayed.

Prefer rebase to merge, and squash commits as needed to preserve a readable
commit history. This project maintains linear history in the `dev` branch, so
we will either rebase or squash your PR when merging. It is much easier for us
if your branch already has a readable commit history (ensure that your commit
subject lines are clear enough to identify the patch in the git log). An
exception to this is made for large PRs that are likely to require multiple
rounds of review; in that case it's easier if you **don't** do this (GitHub
does not preserve the history of old commits, so we cannot filter a PR for only
new changes if a branch is force pushed) and we will squash it when merging.

New features and bug fixes are usually worth mentioning in the changelog.
Exceptions are fixes for bugs that never shipped (were only present in a canary
build), and changes with no intended user observable behavior, such as a
refactor. If you're comfortable writing the note yourself, add it to
`changelog.md` in the root of the project in the section for the upcoming
release.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,7 @@
 * **[UX]** Show an "End of Mission Detected, processing Mission Data" busy dialog while turn results are processed, so the wait is not mistaken for a missed detection
 
 ## Fixes
+* **[Flight Plans]** Fixed IndexError crash when a flight exits combat at its last waypoint
 * **[Mission]** Reliably auto-detect end of mission, even when DCS wrote the final state.json before the wait dialog started watching
 * **[Performance]** Faster post-mission turn processing
 * **[AirWing]** Track per-squadron campaign aircraft stats (initial/destroyed/purchased, save-compatible) and expose pilot experience level and living/dead pilot views for the UI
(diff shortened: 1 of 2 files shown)


Druss99:
If I'm reading this correctly this bug only affects scramble flights which you have added in a separate PR, this commit should be added to that PR instead

PR #779, opened 2026-06-09
Add JAMMING flight type and C-130J EW/ISR mission systems

BradySox:
Adds a new JAMMING flight type that configures the C-130J-30 as an EC-130H Compass Call / RC-130H Rivet Joint electronic warfare and ISR platform.

New FlightType.JAMMING — uses an AWACS-style standoff orbit placed safely behind the threat boundary, Weapon Hold ROE, non-jettison loadout
configure_jamming() assigns the AWACS task with AWACSTaskAction so the aircraft behaves as a proper airborne command/jamming node rather than a transport
c130j_mission_systems.lua (~2,000 lines) is automatically injected into the mission when any JAMMING flights are present. Features:
EW: Area, directional, and spot jamming modes; missile spoofing; energy management; pod loadout selection
ISR: Passive radar detection, two-phase ELINT tracking with map marks, auto-triangulation, SIGINT reports
Crew coordination: handoff briefs to friendly player groups
C-130J-30 unit registered in aircraft YAML so it appears as a selectable airframe
Plugin scaffolding (resources/plugins/c130j/) wired into plugins.json


Druss99:
Mixed up some commits here with scramble logic also added

BradySox:
> Mixed up some commits here with scramble logic also added

Thank you, I’m gonna reupload these when I’m back at my desktop I see the mix matching now

PR #782, opened 2026-06-11
feat(intercept-qra): per-squadron QRA interceptors

### LINKED ISSUE #682: Introduce Intercept Mission type

If a plane is set to mission Intercept, it will wait hot on a parking slot and wait until an enemy airplane is approaching. Once an enemy is in range, the Interceptors will start and attack. Certain faction/doctrines rely way more on Intercept than on Barcap. Moose will do this on its own, given the parameters.

Requirements/ thougts:
Each airfield needs a static object as a warehouse for moose. This could even be included as a new sort of strike target: The Airwing command center.
Moose code generator for this, but thats not really difficult.
Its also not great, if Intercept is used as a sort of dump mission if no other mission is a good fit. A certain number of aircraft should remain on the airfield so its a valid OCA/Aircraft target.
This could also be a first step to generally shift the planner to an Airwing concept. No idea how this would look in detail yet, but at least for OPFOR this could be great. Moose would then take a lot of the responsibility from the autoplaner.

### Additional context

_No response_

geofffranks:
## Intercept (QRA): hot-alert interceptors that scramble on EWR detection

Adds a Quick Reaction Alert system: per-squadron reserve aircraft that sit hot on alert and scramble to intercept inbound raids when the enemy is detected by the radar network — driven by Moose's `AI_A2A_DISPATCHER`, for both coalitions.

### What it does

- **Per-squadron QRA reserve.** A first-class `intercept_reserve` on each BARCAP-capable squadron, **carved out of the plannable inventory** so reserved airframes/pilots aren't double-booked by the ATO. Editable in three places (Air Wing config, Squadron dialog, recruitment menu), propagated wing-wide, and saved with the campaign.
- **Real GCI via Moose EWR integration.** Detection comes from the **same IADS EWR / SAM-as-EWR network Skynet uses** (`dcsRetribution.IADS`), feeding a per-coalition `AI_A2A_DISPATCHER`.
- **Attrition + debrief accounting.** QRA losses are reconciled from in-mission survivors and folded into the debrief loss report, so reserve aircraft shot down on alert are counted like any other loss.

### Why in-air spawns

Cold/hot/runway ground spawns repeatedly left interceptors stuck and unable to taxi on some fields, even when there was no congestion on the taxiways (e.g. Tiyas). In-air spawning (`SetDefaultTakeoffInAir`) is the only method that reliably escapes a congested ground and gets the QRA airborne in time to matter while not affecting any previously planned flights trying to taxi or use the runway.

There was a latent Moose bug that was breaking the in-air spawns, so monkeypatched that to avoid modifying the vendored MOOSE copy. Filed the permanetfix as **[MOOSE #2595](https://github.com/FlightControl-Master/MOOSE/pull/2595)** — the monkeypatch can be dropped once that lands in retribution's copy of Moose.

### Configuration (Campaign Doctrine)

| Setting | Default | Range |
|---|---|---|
| Default QRA reserve per OWNFOR interceptor squadron | 0 | 0–12 |
| Default QRA reserve per OPFOR interceptor squadron | 0 | 0–12 |
| QRA GCI max scramble radius (NM) — gates how close a raid must be before a base scrambles | 100 | 1–400 |
| QRA interceptor engagement range (NM) — how far a scrambled interceptor chases before disengaging (`SetEngageRadius`) | 60 | 1–200 |
| QRA radio scramble callouts (scramble / wheels-up / engaging / RTB) | on | toggle |

Defaults are 0 reserve, so the feature is **off until a player opts in** — existing campaigns are unaffected. Per-squadron values can be tuned after new-game start.

Fixes #682

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # Retribution v1.6.0
 
 ## Features/Improvements
+* **[Mission Generator]** Per-squadron QRA interceptor reserve: OPFOR (and opt-in OWNFOR) squadrons scramble in-air interceptors via the Moose AI_A2A_DISPATCHER against incoming air-to-ground raids detected by the IADS/EWR network. Scramble/engage radius, spawn altitude, grouping, and radio callouts are campaign-doctrine settings.
 * **[Kneeboard]** Use a light-grey daytime kneeboard background instead of near-white, to avoid glare under HDR / Auto-HDR while staying readable in daylight.
 * **[UX]** Hovering a friendly flight's route line on the map highlights it in yellow, and clicking it selects that flight's package (and the flight) in the ATO sidebar.
 * **[UX]** Press Delete with a package selected in the Packages list to cancel it, making it quick to clear several packages in a row.
--- a/game/campaignloader/defaultsquadronassigner.py
+++ b/game/campaignloader/defaultsquadronassigner.py
@@ -5,6 +5,7 @@ import random
 from typing import Optional, TYPE_CHECKING
 
 from game.squadrons import Squadron
+from game.squadrons.intercept_reserve import seeded_intercept_reserve
 from game.squadrons.squadrondef import SquadronDef
 from .campaignairwingconfig import CampaignAirWingConfig, SquadronConfig
 from ..ato.flighttype import FlightType
@@ -49,6 +50,16 @@ class DefaultSquadronAssigner:
                     control_point,
                     self.coalition,
                     self.game,
+                )
+                if self.coalition.player.is_blue:
+                    default_qra_reserve = self.game.settings.ownfor_default_qra_reserve
+                else:
+                    default_qra_reserve = self.game.settings.opfor_default_qra_reserve
+                squadron.intercept_reserve = seeded_intercept_reserve(
+                    squadron.capable_of(FlightType.BARCAP),
+                    squadron.intercept_reserve,
+                    default_qra_reserve,
+                    squadron.max_size,
                 )
                 squadron.set_auto_assignable_mission_types(
                     squadron_config.auto_assignable
--- a/game/campaignloader/squadrondefgenerator.py
+++ b/game/campaignloader/squadrondefgenerator.py
@@ -55,6 +55,7 @@ class SquadronDefGenerator:
             operating_bases=OperatingBases.default_for_aircraft(aircraft),
             female_pilot_percentage=6,
             pilot_pool=[],
+            intercept_reserve=0,
         )
 
     @staticmethod
--- a/game/debriefing.py
+++ b/game/debriefing.py
@@ -167,9 +167,12 @@ class StateData:
             else:
                 killed_ground_units.append(unit)
 
-        intercept_survivors = {
-            str(k): int(v) for k, v in data.get("intercept_survivors", {}).items()
-        }
+        raw_survivors = data.get("intercept_survivors", {})
+        # An empty Lua table serializes to a JSON array ([]) rather than an
+        # object, so coerce any non-dict to an empty mapping before iterating.
+        if not isinstance(raw_survivors, dict):
+            raw_survivors = {}
+        intercept_survivors = {str(k): int(v) for k, v in raw_survivors.items()}
 
         return cls(
             mission_ended=data.get("mission_ended", False),
--- a/game/debriefing.py
+++ b/game/debriefing.py
@@ -18,6 +18,10 @@ from uuid import UUID
 
 from game.dcs.aircrafttype import AircraftType
 from game.dcs.groundunittype import GroundUnitType
+from game.missiongenerator.interceptattrition import (
+    fielded_qra_by_squadron,
+    reconcile_intercept_losses,
+)
 from game.theater import Airfield, ControlPoint, Player
 
 if TYPE_CHECKING:
@@ -127,6 +131,10 @@ class StateData:
     #: Mangled names of bases that were captured during the mission.
     base_capture_events: List[str]
 
+    #: Per-squadron QRA survivor counts reported by the intercept plugin,
+    #: keyed by stringified squadron UUID.
+    intercept_survivors: dict[str, int]
+
     @classmethod
     def from_json(cls, data: Dict[str, Any], unit_map: UnitMap) -> StateData:
         def clean_unit_list(unit_list: List[Any]) -> List[str]:
@@ -159,12 +167,17 @@ class StateData:
             else:
                 killed_ground_units.append(unit)
 
+        intercept_survivors = {
+            str(k): int(v) for k, v in data.get("intercept_survivors", {}).items()
+        }
+
         return cls(
             mission_ended=data.get("mission_ended", False),
             killed_aircraft=killed_aircraft,
             killed_ground_units=killed_ground_units,
             destroyed_statics=data.get("destroyed_objects_positions", []),
             base_capture_events=data.get("base_capture_events", []),
+            intercept_survivors=intercept_survivors,
         )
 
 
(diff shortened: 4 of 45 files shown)


Starfire13:
Was wondering if an option to make them only intercept packages with bombers might make them a little more interesting. Alternatively, only have them intercept packages with certain taskings. Otherwise they'd just get triggered by the first SEAD and TARCAP flights and they would be functionally little different from CAP. Keeping them in reserve to intercept the heavy hitters might be worthwhile.

geofffranks:
I like this idea.  It will also help with situations with tarcap triggering interceptors which trigger interceptors.I'll see what I can do and restrict it to things like bai, strike, oca, air assault, etc. essentially air to ground minus cas?Sent from my iPhoneOn Jul 4, 2026, at 6:12 PM, Starfire13 ***@***.***> wrote:﻿Starfire13 left a comment (dcs-retribution/dcs-retribution#782)
Was wondering if an option to make them only intercept packages with bombers might make them a little more interesting. Alternatively, only have them intercept packages with certain taskings. Otherwise they'd just get triggered by the first SEAD and TARCAP flights and they would be functionally little different from CAP. Keeping them in reserve to intercept the heavy hitters might be worthwhile.

geofffranks:
Ok, it's now spawning intercept on only air-ground threat groups. Right now it's hardcoded. we can make it configurable if desired, or if you think some of them should still trigger intercepts, I can make that change too.

| Mission type (FlightType) | QRA reacts? | Notes |
|---|:---:|---|
| Strike | ✅ | air-to-ground |
| BAI | ✅ | air-to-ground |
| OCA/Runway | ✅ | air-to-ground |
| OCA/Aircraft | ✅ | air-to-ground |
| Anti-ship | ✅ | air-to-ground |
| DEAD | ✅ | air-to-ground (destroy air defense) |
| Armed Recon | ✅ | air-to-ground (added in review) |
| Air Assault | ✅ | air-to-ground (added in review) |
| BARCAP | ❌ | air-to-air CAP |
| TARCAP | ❌ | air-to-air CAP |
| Fighter sweep | ❌ | air-to-air |
| Escort | ❌ | air-to-air |
| Intercept | ❌ | air-to-air |
| SEAD | ❌ | excluded per doctrine (SEAD family) |
| SEAD Escort | ❌ | SEAD family |
| SEAD Sweep | ❌ | SEAD family |
| CAS | ❌ | close air support |
| AEW&C | ❌ | support |
| Refueling | ❌ | support |
| Transport | ❌ | support |
| Cargo Transport | ❌ | support (Pretense) |
| Ferry | ❌ | support |
| Recovery | ❌ | support |

Starfire13:
I'm thinking probably better to leave DEAD out as there will be a DEAD flight in SEAD packages so you will end up with the same issue of intercept triggering on the very first SEAD package.

I also feel like using interceptors to attack an air assault chopper is a bit of an overkill. There are better targets to use intercept on.

geofffranks:
> I'm thinking probably better to leave DEAD out as there will be a DEAD flight in SEAD packages so you will end up with the same issue of intercept triggering on the very first SEAD package.

This is why i'm on the fence about leaving SEAD out. Wouldn't you want to spawn interceptors to prevent your air defenses from being wiped out?

> I also feel like using interceptors to attack an air assault chopper is a bit of an overkill. There are better targets to use intercept on.

Same here - there are better targets, but if this is the only thing standing between your base being taken over, why wouldn't you use it? Though I can see it making air-assault missions much more complicated to ensure escorts are on-station at the right time.

Starfire13:
The way I see it, you only have a single interceptor flight per airfield anyway. So you want the maximum bang for your buck. If you are sending it out after the very first package coming your way (which is likely SEAD/DEAD) then it is going after the same target your BARCAP flight is also going after, and also attacking at a time point where air defence is most intact and capable. This feels like a bit of a waste.

On the other end of the time window, there's the air assault choppers, which typically arrive near the end of the turn when everything is RTBing or dead. It's unlikely that the interceptor flight is still available at that point. I would have expected it to get sent out after the numerous BAI or strike packages that arrive before the chopper does. Also, the choppers are escorted by other choppers, and those escorts are meant to get rid of the occasional ground unit that escaped death after the BAI and Armed Recon flights have gone through. Choppers are not gonna be firing Stingers at an interceptor flying at 20,000 feet.

geofffranks:
ok, those are both ignored now

BradySox:
We've been running this PR's dispatcher in our fork ([414Ret](https://github.com/BradySox/414Ret)) since mid-June across weekly MP events and several campaigns, porting your recent drift back in as it landed (the prefix escape, the A2G react filter, the live reserve accounting, the runway gate — all working well). A few things we changed on top that might be worth folding in, roughly in order of value:

**1. Air-spawn profile: jets spawn at ~0 kt at a fixed MSL altitude.**
Moose's in-air spawn (`SPAWN:SpawnAtAirbase`, `Takeoff.Air`) sets position and altitude but **not speed**, so the spawned group inherits the parking template's ~0 kt — interceptors spawn stalled and dive recovering airspeed. We had a Su-27 dive ~4,600 ft after spawn and nearly hit the ground at Vaziani (Tacview, 2026-06-20). Separately, `SetDefaultTakeoffInAirAltitude(2000)` is absolute MSL for every base, so high-elevation fields eat most of that margin — and at low fields they materialize high with free energy instead of coming off the deck like a scramble. Our fix, per squadron in the `SetSquadron` loop:

```lua
-- terrain-relative spawn altitude (~2,500 ft AGL over the launching field)
local base = AIRBASE:FindByName(rec.airbaseName)
if base then
    local ok_e, elev = pcall(function() return base:GetCoordinate():GetLandHeight() end)
    if ok_e and elev then
        dispatcher:SetSquadronTakeoffInAirAltitude(sq, elev + 760)
    end
end
-- real scramble airspeed (InitSpeed propagates through Moose SpawnWithIndex)
local sq_obj = dispatcher.DefenderSquadrons[sq]
if sq_obj and sq_obj.Spawn then
    for _, sp in ipairs(sq_obj.Spawn) do
        pcall(function() sp:InitSpeedKnots(300) end)
    end
end
```

(keeping the global `SetDefaultTakeoffInAirAltitude(2000)` as a fallback for when the base lookup misses)

**2. Before dropping the backstop EWR entirely — the taxi-routing breakage is fixable with an offset.**
We hit the same problem (the backstop spawns at the airbase reference point, on the runway/taxiway, and breaks AI taxi routing) and fixed it by offsetting the spawn 300 m off the reference point instead of removing it:

```lua
local vec2 = airbase:GetVec2()
local ewr_vec2 = { x = vec2.x + 300, y = vec2.y + 300 }
```

With that (plus the invisible/immortal protection it already had), we've flown backstops for a month-plus with zero taxi blockage. Two reasons to keep it: without the backstop, a coalition whose EWR/SAM-as-EWR net dies loses QRA detection entirely; and — as your own escape-fix commit shows — before that fix the paren-free backstop names were quietly carrying *all* detection, so "EWR network only" is thinner coverage in practice than on paper (some factions/campaigns barely field EWR at all). Could also be a toggle if you'd rather keep network-degradation the default. Happy to PR the offset version into your branch if you want it.

**3. Latent: GCI radius settings above ~162 NM silently abort mid-transit.**
The GCI max radius setting allows up to 400 NM, but Moose aborts a defender once `DistanceFromHomeBase > DisengageRadius` — default 300 km ≈ 162 NM — so a base at the far edge of a wide radius launches and then turns around. Worth coupling `SetDisengageRadius(gci + engage + margin)` whenever the setting exceeds Moose's default, or capping the setting's range at ~160.

**4. Defaults: 100/60 plays as a forward interceptor screen, not QRA.**
Flown experience across several campaigns: with 100 NM scramble + 60 NM engage, every base answers distant raids and winners chase deep into contested airspace. We settled on **60 NM scramble / 38 NM engage** as defaults, and it reads like alert birds defending their own base. (We later built a border-zone "forward defense" mode on top — `SetBorderZone` → per-CP accept zones so rear bases answer the front without the front base chasing past it — which we'd offer as a follow-up PR once this lands rather than grow this one.)

**5. Small QoL bits, take or leave:**
- **Grouping:** instead of a fixed `SetSquadronGrouping(sq, 2)`, we roll per squadron 75% single / 25% pair, emitted per record — most alert launches are singles, so a reserve of 4 answers four raids instead of two.
- **Comms noise:** with comms on, Moose announces "landing at base" for every recovering defender; we wrap `DETECTION_MANAGER:MessageToPlayers` to drop just those (both the English and Russian "посадка на базу" strings), keeping the scramble/engage calls.

geofffranks:
> We've been running this PR's dispatcher in our fork ([414Ret](https://github.com/BradySox/414Ret)) since mid-June across weekly MP events and several campaigns, porting your recent drift back in as it landed (the prefix escape, the A2G react filter, the live reserve accounting, the runway gate — all working well). A few things we changed on top that might be worth folding in, roughly in order of value:

Amazing! thanks for the feedback!


> **1. Air-spawn profile: jets spawn at ~0 kt at a fixed MSL altitude.** 

Yeah I ran into this and it should have been fixed already? Not sure if you were on an old commit or not though. I did incorporate the elev + offset vs raw altitude approach you suggested.

> **2. Before dropping the backstop EWR entirely — the taxi-routing breakage is fixable with an offset.** 

I don't think I want to do this because of the difficulty in making sure the EWR is being offset into the right location. I'm also not sure if i want to have QRA interceptors have to suddenly spawn and taxi and either get stuck waiting for a runway if other flights are taking off/landing at that time, and never get off the ground in time.  What do you think @Starfire13 @Druss99?


> **3. Latent: GCI radius settings above ~162 NM silently abort mid-transit.**

Nice catch, fixed.

> **4. Defaults: 100/60 plays as a forward interceptor screen, not QRA.** 

Updated to your defaults. 

> **5. Small QoL bits, take or leave:**

Did the grouping change, didn't do the comms change since you can turn it off entirely in config (QRA radio scramble callouts)

Starfire13:
I hate air spawns in general as planes magically popping into existence in mid air looks bad, but I don't see how anything else other than that would work smoothly. The risk of things getting fucked up with runway spawns is too high. Either the QRA gets delayed because the runway is in use, or it spawns and stuff crashes into stuff. Could make it a toggleable option in the settings though.

BradySox:
> I hate air spawns in general as planes magically popping into existence in mid air looks bad, but I don't see how anything else other than that would work smoothly. The risk of things getting fucked up with runway spawns is too high. Either the QRA gets delayed because the runway is in use, or it spawns and stuff crashes into stuff. Could make it a toggleable option in the settings though.

Totally agree Air Spawns are cheap, but at-least with what ive seen in tests they are spawning runway threshold, about 500ft, and 300kt. not perfect but that's pretty close to them really coming off the ground

PR #785, opened 2026-06-11
Add map layer toggles to hide destroyed non-repairable ground objects

juanjux:
Add two map overlay toggles, "Red: destroyed (non-repairable)" and
"Blue: destroyed (non-repairable)", in the top-left layer panel (below
the ruler). Turning a side's layer off hides that coalition's fully
destroyed, non-rebuildable ground objects (buildings, ships, missile and
coastal sites). Repairable objects (purchasable: SAM/EWR/armor) are
never hidden, so they remain visible and repairable. Both layers are
shown by default, so the map looks unchanged until toggled.

This helps clear the map, specially on the advanced stages of big campaigns where a very high % of the icons in the map will be destroyed stuff.

Note: this is forward-compatible with #680 , it will detect if the stuff is repairable or not instead of hardcoding it.


juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/16

PR #789, opened 2026-06-13
Fix inverted OPFOR auto-planner aggressiveness roll

BradySox:
## Problem

`opfor_autoplanner_aggressiveness` is documented in `settings.py` — and used by `PackagePlanningTask._get_weighted_threat_range` (`margin = 100 - aggressiveness`) — as the percentage of the threat radius the OPFOR autoplanner ignores: **0 considers every threat (always defend), 100 ignores them all (always commit offensively).**

But `ObjectiveFinder.vulnerable_control_points()` rolls `randint(1, 100) > aggressiveness` to zero out a base's threat range and plan offensively. That's backwards:

- At **100%** (max aggressive): `randint(1,100) > 100` is never true → it *never* plans offensively.
- At **0%** (max cautious): `randint(1,100) > 0` is always true → it *always* plans offensively.

## Fix

Use `<=` so the chance of planning offensively rises with the setting, consistent with the description and `_get_weighted_threat_range`. The roll is extracted into `_opfor_plans_offensively()` so the intended semantics are documented and unit-testable.

## Tests

Adds `tests/test_objectivefinder_aggressiveness.py` pinning the boundary behavior (0 never, 100 always, threshold inclusive, BLUE unaffected). Black- and mypy-clean.

### DIFF
--- a/game/commander/objectivefinder.py
+++ b/game/commander/objectivefinder.py
@@ -158,7 +158,7 @@ class ObjectiveFinder:
             if (
                 self.is_player.is_red
                 and randint(1, 100)
-                > self.game.settings.opfor_autoplanner_aggressiveness
+                < self.game.settings.opfor_autoplanner_aggressiveness
             ):
                 # Chance that the airfield threat range will be evaluated as zero,
                 # causing the OPFOR autoplanner to plan offensively


game/commander/objectivefinder.py
```diff
@@ -140,6 +140,21 @@ def front_lines(self) -> Iterator[FrontLine]:
         """Iterates over all active front lines in the theater."""
         yield from self.game.theater.conflicts()
 
+    def _opfor_plans_offensively(self) -> bool:
```
Druss99: This doesn't need it's own method, we can just swap the > to a < in the existing method

Druss99 (changes requested):

Druss99 (approved):

PR #793, opened 2026-06-15
Don't show "Missing Recon Picture" placeholder on building cards

BradySox:
## Problem

`SceneryUnit.icon` always returns `"missing"` (`game/theater/theatergroup.py`), so every scenery-building card in the ground-object menu loads `resources/ui/units/buildings/missing.png` as its header image — and that PNG contains the literal text **"Missing Recon Picture"**. The result is a large cyan placeholder block on every building card where no real icon exists.

## Fix

`QBuildingInfo.init_ui()` now skips the image widget entirely when there is no real icon (icon is `"missing"` or the file is absent) and the building is alive, leaving a compact, top-aligned **name + value** card. Dead buildings still show `dead.png`, and buildings that *do* have a real icon still show it.

Also: word-wrap the name label and remove a leftover unused `QHBoxLayout` footer.

## Notes

UI-only, no behavioral change to mission/economy logic. Black clean.

### DIFF
--- a/qt_ui/windows/groundobject/QBuildingInfo.py
+++ b/qt_ui/windows/groundobject/QBuildingInfo.py
@@ -1,7 +1,8 @@
 import os
 
+from PySide6.QtCore import Qt
 from PySide6.QtGui import QPixmap
-from PySide6.QtWidgets import QGroupBox, QHBoxLayout, QLabel, QVBoxLayout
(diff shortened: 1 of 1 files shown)


juanjux:
Been testing this for a while on my branch, it works perfectly.

juanjux (approved):

PR #794, opened 2026-06-15
Hide mobile SAM/AAA/MANPAD units riding inside non-air-defense groups on the MFD

BradySox:
## Problem

`hidden_on_mfd` is a **group-level** DCS property, set in `GroundObjectGenerator` from the parent ground object's `hide_on_mfd` flag. A SHORAD / AAA / MANPAD escort that is generated *inside* an armor or missile group therefore inherits that group's *visible* flag and stays on the datalink / HSI — betraying its position — even though the same unit standing alone (in an air-defense group) would be hidden.

## Fix

`create_vehicle_group()` / `create_ship_group()` now OR the existing flag with a new `_contains_mobile_air_defense()` check: a generated group is hidden when it **contains** a unit of class `AAA`, `SHORAD`, or `MANPAD` (new `MOBILE_AIR_DEFENSE_UNIT_CLASSES` in `game/data/units.py`).

`TELAR` and the radar/launcher classes are deliberately **excluded**, so standalone mobile MERAD/LORAD SAM sites (SA-6/11, SA-2/3/5/10) stay visible and targetable for SEAD.

## Tests

`tests/missiongenerator/test_mobile_air_defense_hiding.py` — escort detection, pure-armor negative case, and the TELAR-stays-visible case. Black + mypy clean.


juanjux:
This probably should be gated behind an option. I could imagine mobile AA defenses in the real world could be added to pilot's MFDs if identified in recon or recent satellite images.

BradySox:
> This probably should be gated behind an option. I could imagine mobile AA defenses in the real world could be added to pilot's MFDs if identified in recon or recent satellite images.

Maybe roll it into the recon work? The hiding sams is a what caused my very first commit I think the hiding MOBILE sams is a real addition, the hiding everything else could roll into the BDA/Recon fog?

juanjux:
> > This probably should be gated behind an option. I could imagine mobile AA defenses in the real world could be added to pilot's MFDs if identified in recon or recent satellite images.
> 
> Maybe roll it into the recon work? The hiding sams is a what caused my very first commit I think the hiding MOBILE sams is a real addition, the hiding everything else could roll into the BDA/Recon fog?

@bradyccox yes, that could make sense. Alternatively, the position could be the one in the previous turn, which maybe is not the one in the current.

Ramius007:
> This probably should be gated behind an option. I could imagine mobile AA defenses in the real world could be added to pilot's MFDs if identified in recon or recent satellite images.

We currently dont have modules in DCS, that allow RT data sharing with satelites. Even in 2026 decision loop is a real thing.

PR #795, opened 2026-06-16
Fix app process lingering in the background after the window is closed

juanjux:
**Problem:** Closing Retribution sometimes left `retribution_main.exe` running in the background (~800 MB, sometimes I've seen 1500MB!).

The API server (uvicorn) runs in a background thread; on shutdown `Server.run_in_thread()` joins it. But `timeout_graceful_shutdown` was never configured, so uvicorn waited forever for the long-lived `/eventstream` websocket task to drain — it never does when the web view is torn down on close — and the join hung.

**Fix:** Set `timeout_graceful_shutdown=3` on the uvicorn config, plus a `force_exit` + bounded-join backstop, so the server thread always stops and the process exits.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -24,6 +24,7 @@
 * **[UI]** Avoid a crash dialog ("'QWidgetItem' object has no attribute 'width'") when a list using the two-column row delegate relayouts with a malformed style option under PySide6 6.4.x.
 
 ## Fixes
+* **[App]** Fix Retribution sometimes staying alive in the background after the window is closed.
 * **[AirWing]** Squadron transfer-destination parking now accounts for already-ordered incoming transfers, matching the Airfield Command hangar count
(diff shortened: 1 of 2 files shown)


Starfire13:
I've found this also happens if you run more than one instance of the app at the same time. The second (and subsequent) instances do not show up on the screen but show up in task manager, and they do not get automatically shut down when the first instance is terminated.

juanjux:
> I've found this also happens if you run more than one instance of the app at the same time. The second (and subsequent) instances do not show up on the screen but show up in task manager, and they do not get automatically shut down when the first instance is terminated.

The specific case handled by this PR happened only with a single instance, I tested opening more instances on my patched fork and it still happens as you say so this PR doesn't fix that case, I will take a look at that other one since it could be potentially easy to fix with some kind of lock of Windows property.

juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/15

PR #796, opened 2026-06-16
Write loadout payload files atomically (avoid half-written files)

juanjux:
Saving a loadout in the payload editor wrote the `.lua` in place, leaving a window where a concurrent reader (mission generation, the loadout editor) could read a truncated, half-written file. This writes to a temp file in the same directory and `os.replace()`s it into place, so a reader only ever sees the old or the new file — never a partial one.

(The parse-error handling that previously lived here is superseded by #751, which keeps the payloads it can read instead of dropping the whole airframe.)


juanjux:
Consolidated into #751, which now carries both the read-side fix (skip unparseable payload files) and the atomic loadout writes from this PR. Closing in favour of #751.

PR #797, opened 2026-06-16
Fix kneeboard waypoint numbering for in-air-start flights

juanjux:
**Problem:** For flights that start in the air, DCS spawns the group at its current waypoint (group point 0) and omits the already-passed takeoff/join/ingress waypoints, so the in-game waypoint numbers start later in the plan. The kneeboard was built from the full flight plan starting at takeoff, so its numbering and labels were offset from the cockpit (e.g. kneeboard "1: Hold" while DCS showed "1: Escort Hold").

**Fix:** `create_waypoints()` now returns the plan from the spawn waypoint onward for `InFlight` states, so the kneeboard flight-plan page and strike-task page match the generated mission. Ground/runway starts are unchanged.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -24,6 +24,7 @@
 * **[UI]** Avoid a crash dialog ("'QWidgetItem' object has no attribute 'width'") when a list using the two-column row delegate relayouts with a malformed style option under PySide6 6.4.x.
 
 ## Fixes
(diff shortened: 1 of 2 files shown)


juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/14

PR #800, opened 2026-06-17
Mirror COMM1 radio presets onto COMM2 on aircraft with a second radio

juanjux:
Retribution assigns every preset radio channel (departure/arrival ATC, AWACS, JTAC, package, tankers, divert) to the inter-flight radio only. On a two-radio aircraft like the F/A-18C or F-16 you therefore can't keep, for example, your package on COMM1 and AWACS on COMM2 at the same time -- which is how manual campaign missions are usually set up.

This mirrors the same preset set onto the second radio (the intra-flight radio, which is COMM2 on the Hornet/Viper), starting at channel 2 since channel 1 carries the intra-flight frequency. It is driven by each aircraft's actual radio layout, so single-radio types and JF-17-style shared-preset radios are skipped automatically. The kneeboard now shows every channel a frequency maps to (e.g. `COMM1 Ch 3 / COMM2 Ch 4`).

The Tankers/JTAC kneeboard tables were also tightened so the wider FREQ column still fits (drop the always-"Tanker" Task column, `TACAN`->`TCN`, `Laser Code`->`Laser`).

Support page kneeboard

Mission info kneeboard

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -16,6 +16,8 @@
 * **[Map]** Hovering a SAM threat or detection ring highlights its emitter — and hovering an emitter highlights its ring — making it easy to tell which site a ring belongs to. Can be disabled from the map's layer control.
 
 ## Fixes
+* **[Comms]** Fix COMM1/COMM2 being swapped in the kneeboard for the Super Hornet family (EA-18G, F/A-18E/F and their tanker variants): align their radio indices to the F/A-18C so intra-flight is COMM2 and ATC/package are on COMM1.
+* **[Kneeboard]** Stop the Support Info FREQ column being clipped now that it lists both COMM1 and COMM2 (the single-digit aircraft-count column no longer pads the row past the page edge).
 * **[Mission]** Reliably auto-detect end of mission, even when DCS wrote the final state.json before the wait dialog started watching
 * **[Performance]** Faster post-mission turn processing
 * **[AirWing]** Track per-squadron campaign aircraft stats (initial/destroyed/purchased, save-compatible) and expose pilot experience level and living/dead pilot views for the UI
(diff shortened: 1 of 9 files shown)


Starfire13:
How would this impact the aircraft that have, say intraflight on radio 1 and interflight on radio 2 like the Hornet,rather than everything on one radio? The aircraft I've flown mostly use the two radios for separate things, as far as I can recall. For example, Viper has radio 1 for interflight and radio 2 for intraflight.

juanjux:
> How would this impact the aircraft that have, say intraflight on radio 1 and interflight on radio 2 like the Hornet,rather than everything on one radio? The aircraft I've flown mostly use the two radios for separate things, as far as I can recall. For example, Viper has radio 1 for interflight and radio 2 for intraflight.

@Starfire13 it will like that by default for planes with more than one radio, radio 1 for interflight (comm 1 1 by default, connected to ATC) and radio 2 for intraflight (comm 2 1 which is assigned to the flight), of course the user can change it at any time as everything will be available in both comms.

Starfire13:
How is that different from what we have now, where one radio is for interflight and one is for intraflight?

juanjux:
> How is that different from what we have now, where one radio is for interflight and one is for intraflight?

The difference is that you can tune any two targets on any of the two radios, you are not limited to always have COMM2 for intraflight and COMM1 for everything else, e.g. you could have package radio on COMM2 (which include the interflight: orders to the wingmen work) and ATC/AWACS/tanker/whatever on COMM1. Or the other way around. Or AWACS and ATC (e.g. if you are flying alone without wingmen, right now COMM2 is wasted on nothing) or... whatever you want.

However, by default, it will be like before the PR, COMM2 on intraflight and COMM1 for ATC of the launch airport.

juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/12

PR #754, opened 2026-05-25
Adds Additional Kneeboard Generation with Maps/Flight plans for player packages

geofffranks:
Adds target reconnaissance kneeboard pages to mission generation, plus QFE/field-elevation info on the kneeboards. Two commits on upstream/dev: the feature code/tests/docs, and the supporting airport-imagery dataset.

1. Target recon kneeboard pages — new per-flight kneeboard pages for player air-to-ground flights, rendered over an Esri World Imagery satellite basemap: an airfield-departure page (spawn/runway/ATIS), an overview + detail page for strike targets, an airbase page (OCA), and a front-line detail page (CAS). Includes threat rings, aimpoint/unit markers, label-collision avoidance, and MGRS coordinates.

2. QFE from field elevation — reduces sea-level QNH to airfield-level QFE using OSM/DEM-derived field elevation, surfaced both on the standard briefing page and the departure kneeboard ATIS block.

## Backwards compatibility
Additive. Recon pages are new pages appended for eligible flights; existing kneeboard pages are unchanged.
Graceful degradation. If the Esri tile service is unreachable, the basemap falls back to a GIF/landmap renderer with an OFFLINE banner (the old path still works). If no elevation is shipped for an airport, the QFE line is simply omitted (briefing) or shows N/A (ATIS) — no errors, QNH behavior unchanged.
One new dependency: mgrs>=1.4.6. Esri imagery needs no API key.

## Opt in/out
Recon pages: Mission Generator → Kneeboard → "Generate target recon kneeboard pages", default ON. Set false to disable entirely. Tuning knob: "Extra threat search radius (nmi)" (target_recon_extra_threat_search_nmi, default 0). Pages only emit for player flights with A/G tasks regardless.

## Ongoing maintenance
Airport-imagery dataset (resources/airport_imagery/<terrain>.json): regenerate when DCS adds/moves airports or OSM corrects runway/elevation data, via scripts/derive_airport_imagery_offsets.py (OSM Overpass + Open-Elevation DEM). Cadence is loose — see resources/airport_imagery/README.md. Missing data just disables imagery offset + QFE for that airfield.
External services: Esri World Imagery (not a new dependency - same service as what generates the main screen's map data) and Open-Elevation (only needed when running the airport runway/elevation dataset-generation script).

[...]

PR #801, opened 2026-06-17
kneeboard's friendly-packages list + package-targets map pages

juanjux:
Adds two optional kneeboard pages, appended after the existing ones, toggled by **Mission Generator → Kneeboard → "Generate all-packages kneeboard pages"** (default on):

1. **Friendly Packages** — one row per friendly package with its TOT (patrols show their on-station window instead), so flights can deconflict timing across the ATO. 
2. **Package Targets Map** — a filled-coastline theater overview (built from the recon module's landmap, which is why this PR depends on #754) labelling where every attack package is headed (Strike / CAS / DEAD / SEAD / BAI / anti-ship / OCA / air-assault / recon — not the support patrols).

## ⚠️ Depends on #754

Once #754 lands I'll rebase onto `dev` so only the three kneeboard commits remain:

- `feat(kneeboard): add a friendly-packages coordination page`
- `feat(kneeboard): add a package-targets map page after the packages list`
- `kneeboard: bigger packages-list font; coastline theater map for package targets`


juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/11

PR #803, opened 2026-06-18
Colour enemy IADS network links by kind and state, and ease the tooltip hover

juanjux:
## What

When the **Enemy IADS Network** overlay is enabled, every connection line was drawn in a single washed-out colour, so you couldn't tell a live link from a dead one, and the line was a pixel-thin hit target that made the status tooltip hard to hover.

This colours each link by both **kind** and **state**, and gives the tooltip a comfortable hover margin:

| Link | Active | Inactive |
| --- | --- | --- |
| Communication | green | red |
| Power | cyan | gold |

Inactive links are also drawn dashed (and slightly more transparent) so the state reads at a glance even for the colour-blind.

The two power tones are deliberately picked to stay distinct from the existing map layers: the cyan avoids the routes / threat-zone blue (`#0084ff`), and the gold avoids the SAM detection-range yellow (`#eee17b`) and the selection highlight (`#ffff00`).

## Hover margin

The visible line is still thin (`weight=2`). Hovering is handled by a second, invisible (`opacity=0`) `weight=16` polyline laid over the same points that carries the sticky tooltip — Leaflet interactive paths use `pointer-events: auto`, so the zero-opacity line still captures the mouse. You no longer need a pixel-perfect hit on the thin line to read the **Active/Inactive** status.

## Scope

Client-only — a single file, `client/src/components/iadsnetwork/IadsNetwork.tsx`. No API, backend, or schema changes.

### DIFF
--- a/client/src/components/iadsnetwork/IadsNetwork.tsx
+++ b/client/src/components/iadsnetwork/IadsNetwork.tsx
@@ -3,37 +3,67 @@ import { Polyline as LPolyline } from "leaflet";
 import { useRef } from "react";
 import { Polyline, Tooltip } from "react-leaflet";
 
+// Colour each IADS link by both kind and state, with tones picked to stay
+// distinct from the other map layers:
+//   Communication: green = active, red  = inactive
+//   Power:         cyan  = active, gold = inactive
+// Power's cyan avoids the routes / threat-zone blue (#0084ff), and its gold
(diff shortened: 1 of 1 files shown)


juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/10

PR #804, opened 2026-06-18
Anubis C-130 Hercules: suppress_ballute crash + air-assault zig-zag route (C-130 & helos)

juanjux:
Two fixes: a crash workaround for the Anubis C-130 Hercules mod, and the air-assault zig-zag ingress route (for the C-130 **and** helicopters).

## 1. Workaround the `suppress_ballute` crash (DCS closes)

The Anubis C-130 airdrop weapons are class `wAmmunitionBallute`. Since the official C-130J sound update, dropping one (or having the carrying C-130 destroyed) makes DCS resolve a sound scheme that has no `suppress_ballute` input port, throws **`No input port in scheme: suppress_ballute`**, and the sim closes to desktop. The common trigger in Retribution is `Herc_Soldier_Squad`, the air-assault paratroop drop.

Fix: strip `Herc_Soldier_Squad` when equipping pylons (`flightgroupconfigurator.setup_payload`), so even loadouts already stored in a save never carry it — editing the payload `.lua` alone only affects newly-planned flights. The weapon is also removed from the customized Hercules payloads.

- **Air-assault base capture is unaffected**: the troops come from CTLD (`cabin_size` + target zone), not the cosmetic soldier-squad pylon.
- **The MOAB (`Herc_GBU-43/B(MOAB)`) is deliberately kept** — it's a usable weapon and a much rarer trigger (documented residual risk in the code comment). The AC-130 gunship loadout is untouched.

Paratroopers in air assaults still work as before.

## 2. Fix the air-assault zig-zag ingress route (C-130 + helicopters)

The Hercules air-assault used `package.waypoints.initial` for its ingress waypoint while the join used `package.waypoints.ingress`, so the route ran out to one point, back to the other, then to the target — a visible approach / retreat / approach zig-zag. Now the ingress waypoint uses `.ingress` too, so join and ingress coincide and the route runs straight to the target.

Helicopters had the same zig-zag for air assault from a different cause: `WaypointBuilder.ingress` re-anchors the helo ingress waypoint to 5 NM from the target, but the join was still built from the raw `package.waypoints.ingress` (the IP, typically 10–45 NM out), so the helo flew in, back out to the IP, then in again. The join now uses `ingress.position`, so it coincides with the (helo-adjusted) ingress and the route runs straight in.

### DIFF
--- a/game/ato/flightplans/airassault.py
+++ b/game/ato/flightplans/airassault.py
@@ -125,18 +125,15 @@ class Builder(FormationAttackBuilder[AirAssaultFlightPlan, AirAssaultLayout]):
             pickup.alt = altitude
             pickup_position = pickup.position
 
-        ingress = (
-            builder.ingress(
-                FlightWaypointType.INGRESS_AIR_ASSAULT,
-                self.package.waypoints.ingress,
-                self.package.target,
-            )
-            if not self.flight.is_hercules
-            else builder.ingress(
-                FlightWaypointType.INGRESS_AIR_ASSAULT,
-                self.package.waypoints.initial,
-                self.package.target,
-            )
+        # Use the package ingress point for the ingress waypoint -- the same point
+        # the join uses below. The Hercules previously ingressed via
+        # package.waypoints.initial while the join stayed at .ingress, so its route
+        # ran out to one point, back to the other, then to the target (a visible
+        # zig-zag). Helicopters never had this because both already used .ingress.
+        ingress = builder.ingress(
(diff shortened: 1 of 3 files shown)


juanjux:
@Starfire13 I think you could be interested in this since while googling what could be happening I saw you having the same issue in the DCS forums and writing that you had to uninstall the AC130 mod.

juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/9

PR #805, opened 2026-06-18
UI: bulk-set flight altitude and step waypoint alt editor by 1000 ft

BradySox:
## Problem

Changing a flight's cruise altitude in the Edit Flight → Waypoints tab means editing every waypoint's **Alt (ft)** cell one at a time. On top of that, the cell's spin box uses the `QDoubleSpinBox` default step of **1 ft**, so the up/down arrows are effectively unusable for real altitude changes — you have to type every value.

## Changes

- **Bulk altitude setter.** Adds an *Altitude* block at the top of the Waypoints tab's action column: a 1000-ft-step spin box plus an **Apply to all** button that writes a single MSL altitude onto every en-route waypoint at once. The spinner seeds from the highest en-route altitude already planned.
  - Takeoff, landing, descent/pattern, AGL (`RADIO`) and Bullseye waypoints are left untouched, so only the cruise/patrol/ingress legs move.
  - Per-waypoint editing is unchanged, so a low-level ingress leg can still be hand-tuned after a bulk set.
- **1000-ft step on the per-waypoint editor.** The `Alt (ft)` cell editor now steps by 1000 ft and shows whole feet (no decimals).

## Notes

UI-only change; no save-format or planner changes. AI flight plans are unaffected. Manually tested in the flight editor on Syria.

### DIFF
--- a/qt_ui/windows/mission/flight/waypoints/QFlightWaypointTab.py
+++ b/qt_ui/windows/mission/flight/waypoints/QFlightWaypointTab.py
@@ -5,9 +5,11 @@ from PySide6.QtCore import Signal, Qt, QModelIndex
 from PySide6.QtWidgets import (
     QFrame,
     QGridLayout,
+    QHBoxLayout,
     QLabel,
     QMessageBox,
     QPushButton,
+    QSpinBox,
     QVBoxLayout,
     QWidget,
 )
@@ -20,6 +22,8 @@ from game.ato.flightplans.planningerror import PlanningError
 from game.ato.flightplans.waypointbuilder import WaypointBuilder
 from game.ato.flighttype import FlightType
 from game.ato.flightwaypoint import FlightWaypoint
+from game.ato.flightwaypointtype import FlightWaypointType
+from game.utils import feet
 from game.ato.loadouts import Loadout
 from game.ato.package import Package
 from game.theater import Player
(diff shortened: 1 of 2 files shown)


qt_ui/windows/mission/flight/waypoints/QFlightWaypointTab.py
```diff
@@ -34,6 +38,17 @@
...
     loadout_changed = Signal()
 
+    # Waypoint types whose altitude the bulk setter must not touch: ground/pattern
+    # points that are tied to the airfield rather than the en-route cruise band.
+    BULK_ALTITUDE_SKIP_TYPES = frozenset(
+        {
+            FlightWaypointType.TAKEOFF,
+            FlightWaypointType.LANDING_POINT,
+            FlightWaypointType.DESCENT_POINT,
+            FlightWaypointType.BULLSEYE,
```
Druss99, line 64: I think there are some waypoint types missing here like TARGET_POINT, DIVERT, etc.

Druss99 (changes requested):

BradySox:
@Druss99 added TARGET_POINT, DIVERT and the other target/landing-zone types to the skip list in the 06-25 commit.

Druss99 (approved):

PR #806, opened 2026-06-18
Configurable cruise/patrol altitude: scatter band + patrol floor

BradySox:
## Problem

Cruise/patrol altitude is effectively hardcoded:

- The per-flight altitude randomization is a single symmetric offset (`randint(0, max) * ±1`) — it can't be biased high/low or turned off.
- There's no way to keep CAP above each airframe's speed-derived preferred altitude, so e.g. an F/A-18C patrols around 24,000 ft with no campaign-level lever to raise it.

This also addresses the existing `# TODO: somehow make ... configurable` notes on the preferred-altitude helpers in `aircrafttype.py`.

## Changes

Two **Campaign Doctrine** settings, both defaulting to **no behavior change** (existing saves are unaffected until opted into):

- **Altitude scatter band.** Replaces the single `max_plane_altitude_offset` with a `[min_plane_altitude_offset, max_plane_altitude_offset]` band (×1000 ft). Defaults `-2`/`+2` reproduce today's ±2k spread. Equal bounds disable scatter (`0`/`0` = none); an asymmetric band biases it (`0`/`+4` = climb-only). The roll is the pure helper `roll_plane_altitude_offset(low, high)`.
- **Minimum patrol altitude** (`min_patrol_altitude`, default `0` = off). Floors CAP/patrol legs to at least the configured altitude *after* the offset, capped by the doctrine's `max_combat_altitude`. Flights already planned higher are unchanged; helos are exempt. Pure helper `apply_patrol_altitude_floor()`, called from `WaypointBuilder.get_patrol_altitude`.

On old saves, `Settings.__setstate__` mirrors the legacy symmetric `max_plane_altitude_offset` into the new minimum (`min = -max`), so every existing save keeps its exact band — including `max = 0` (scatter deliberately off) and widened bands like `max = 5`. Both fields render on the Campaign Doctrine page with no extra UI wiring.

## Tests

`tests/test_flight_altitude_settings.py` — defaults preserve the legacy band, the legacy-save mirror migration (max 0/2/5, and a modern save is never re-mirrored), settings are user-visible, scatter band bounds/equal-bounds/swapped-bounds, and the patrol floor (off / raises low / leaves high / capped by ceiling / never lowers). `black`, `mypy game`, and the new tests pass locally.


Druss99:
I think before we make this change we should update the aircraft yaml files and add some better patrol_altitude, cruise_altitude attributes and see how that performs. These are already configurable in those files but they are optional and no aircraft have them

BradySox:
Closing in favor of Druss99's suggestion above — authoring real patrol_altitude values directly in each fighter's yaml instead of a global settings band. Opening that as a follow-up PR.

BradySox:
Follow-up: #925

PR #808, opened 2026-06-18
Map tooltips: SAM rings (name + emitters) and package routes (flight/package info)

juanjux:
## What

Two map hover tooltips, so you can read what's on the map without clicking through.

**SAM / air-defense rings** — hovering a threat or detection ring shows the site's name (its codename) and a compact, de-duplicated list of the units that make it up, e.g.:

> **AUROCHS**
> 2x SA-6 TEL
> Straight Flush STR

**Package routes** — hovering a friendly route line shows the flight's callsign or composition, its task, the package target and the time-over-target, e.g.:

> **2x F/A-18C Hornet** - CAS
> Target: AUROCHS (TOT 14:16:32Z)

## Notes

- The route tooltip needs information the `Flight` API model didn't carry, so `FlightJs` now also serialises `aircraft`, `num_aircraft`, `flight_type`, `callsign`, `package_target` and `package_tot`. The corresponding fields were hand-added to the generated client `Flight` type (optional, so they degrade gracefully); they'll be reproduced next time the API types are regenerated. The SAM tooltip needs no backend change — it reuses `name` + `units` already on the TGO.

- The route line was only as hoverable as its ~3 px visible width. It now keeps the thin visible line non-interactive and lays a wide, invisible overlay polyline over it that carries the hover (yellow highlight + tooltip) and the click-to-select — the route looks identical but is far easier to grab, mirroring the SAM rings' invisible hover band.

### DIFF
--- a/client/src/api/_liberationApi.ts
+++ b/client/src/api/_liberationApi.ts
@@ -404,6 +404,12 @@ export type Flight = {
   position?: LatLng;
   sidc: string;
   waypoints?: Waypoint[];
+  aircraft?: string;
+  num_aircraft?: number;
+  flight_type?: string;
+  callsign?: string | null;
+  package_target?: string;
+  package_tot?: string;
 };
 export type FrontLine = {
   id: string;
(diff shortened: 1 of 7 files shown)


juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/8

PR #809, opened 2026-06-18
Finances dialog: show income, automated HQ spending and net per turn

juanjux:
Reworks the **Finances** dialog so the gap between income and the balance that actually lands is explicit: gross income, the HQ's automated per-category spending, and the resulting net per turn. Income breaks down per base and per (collapsible) building category, and the automation toggles can be flipped from the dialog.

Finances dialog

The auto-spend rows adapt to the build: front line, aircraft and runway repairs out of the box; building / ground-object repair rows appear on their own if that automation is present. `ProcurementAi.spend_budget` records per-category spend, `Coalition` exposes it as `last_turn_expenses`. The compact `FinancesLayout` used by the intel panel is unchanged.


juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/7

PR #810, opened 2026-06-19
Map: show carrier/LHA ship group, escort threat rings and partial-damage status

juanjux:
Carrier/LHA control points now surface their ship group on the map the way ordinary naval groups do. The carrier and its escorts are carried as an `is_control_point` ground object that is not emitted as a standalone TGO, so until now a battered carrier group looked defenceless on the map and showed no detail as ships were lost.

- The control-point tooltip lists the carrier and its escorts (lost ships keep the `[DEAD]` tag).
- The surviving escorts' air-defense threat and detection rings are drawn on the map.
- The status marker stays yellow while any ship in the group survives, turning red only once the whole group is gone (previously red as soon as the carrier itself sank). Cosmetic only — `runway_is_operational()` still governs launches.

`ControlPointJs` gains `units`, `threat_ranges` and `detection_ranges`; the SAM-only `TgoRangeCircles` is generalised to `RangeCircles` and reused for both TGOs and these control points, so the carrier rings get the same hover-highlight as SAM rings.


juanjux:
Closing as a duplicate of #748, which already implements this feature and is kept up to date with dev.

PR #812, opened 2026-06-19
Bulk-set waypoint altitude and step the altitude editor by 1000 ft

juanjux:
Withdrawn.


juanjux:

PR #813, opened 2026-06-19
Route AI around the active front line

juanjux:
Withdrawn.


juanjux:

PR #814, opened 2026-06-19
Spread frontline units along the line instead of stacking on one tile

juanjux:
Withdrawn.


juanjux:

PR #816, opened 2026-06-19
Fix aircraft sale exploit that corrupted squadron counts

juanjux:
### The bug

Selling aircraft was exploitable. Anything that makes the game re-plan the turn — selling, buying or repairing a unit on the map, or a cheat capture — handed the just-sold aircraft back as if they had never been sold. So you could:

- **Sell the same aircraft over and over**, getting refunded their full price each time → effectively unlimited money.
- **Still fly the "sold" aircraft** on missions that same turn.

Selling more aircraft than a squadron actually owns also left it with a *negative* aircraft count at the end of the turn. On the next load the save-repair step turned that negative count into a number close to the airfield's entire parking capacity — e.g. a 6-aircraft B-52 squadron reappearing with ~994 aircraft.

### The fix

- Sold aircraft stay sold when the turn re-plans: they can no longer be re-sold or flown, and selling refunds their price exactly once.
- The save-repair step no longer inflates a corrupted (negative) count into a near-full airfield — it resets it to zero, while still trimming a genuinely overcrowded base down to what its parking can hold.
- Campaigns already corrupted by this bug still load.

Includes a regression test; full test suite passes.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -15,6 +15,7 @@
 * **[Map]** Hovering a SAM threat or detection ring highlights its emitter — and hovering an emitter highlights its ring — making it easy to tell which site a ring belongs to. Can be disabled from the map's layer control.
 
 ## Fixes
+* **[AirWing]** Selling aircraft no longer lets the same units be re-sold (and re-flown) after a turn re-initialisation, which refunded their price repeatedly and could drive a squadron's aircraft count negative and then balloon it to ~airfield capacity on reload.
 * **[Mission]** Reliably auto-detect end of mission, even when DCS wrote the final state.json before the wait dialog started watching
 * **[Performance]** Faster post-mission turn processing
 * **[AirWing]** Track per-squadron campaign aircraft stats (initial/destroyed/purchased, save-compatible) and expose pilot experience level and living/dead pilot views for the UI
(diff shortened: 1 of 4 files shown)


Starfire13:
IIRC, this was previously picked up on but was left as is as a sneaky way to allow players give money to opfor. I don't think "exploiting money" is an issue considering we already have a built in way to get as much money as you want via cheats (but this method is currently the only way to give opfor money).

I don't really mind if it does get fixed, but we might want to consider adding in a cheat to give opfor money if so.

juanjux:
I also didn't mind the cheat (in a single player game you should be free to do whatever you want) but  the 994 planes on the next turn was definitely an issue, the planner doesn't forbit you to use sold planes so it's not trivial to cause this by accident.

I will add another PR (another one to the pile, sorry 🤣 ) to give or take money to opfor in the same UI of the money cheat, shoud be pretty easy.

juanjux:
@Starfire13 your wishes are orders 🧞‍♂️  https://github.com/dcs-retribution/dcs-retribution/pull/818

juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/5

PR #817, opened 2026-06-19
Let non-DEAD-role aircraft fly DEAD manually (secondary tasks)

juanjux:
Some aircraft can carry a good standoff/PGM weapon but DEAD is not one of their natural roles (A-10, F-35, Su-57, Su-35S, F-117A, strategic bombers…). Auto-assigning them to DEAD would be wrong, but a player may still want to task them with it manually.

**Mechanism:** aircraft yamls gain an optional `secondary_tasks` list. A task listed there is a real capability — `capable_of()` is true, it shows in the UI and can be assigned manually — but it is left out of a squadron's **default** auto-assignable mission types, so its checkbox starts **unchecked** (the AI won't auto-plan it; the player can tick it or assign manually). `capable_of` / `iter_task_capabilities` are unchanged; only the default auto-assignable set built by the squadron-def generators subtracts the secondary tasks.

DEAD is marked secondary for these airframes, each given a valid, sensible DEAD loadout (best standoff PGM it can field, preferring fire-and-forget GPS/INS over TV/laser; no anti-radar, decoys or cluster bombs):

| Aircraft | Weapon |
|---|---|
| A-10C / A-10C_2, MQ-9 Reaper | JDAM (GPS) |
| A-10A, F-16A, F-35C | Maverick (imaging-IR) |
| F-35A / F-35B | AGM-84E SLAM |
| Su-57, Mirage-F1EQ | Kh-29T (TV) |
| Su-35S | Kh-59MK2 |
| F-117A | GBU-10 (built-in designation) |
| Su-25 | Kh-29L / Kh-25ML |
| Tu-95MS, CH_Tu-160M2 | Kh-101 cruise |
| Tu-160 | Kh-65 cruise |
| Tu-16, Tu-4K, Tu_95K, tu_22D, tu_22KD | (kept their existing DEAD task, now secondary — only dumb bombs, no preset) |

Left out: M-2000C, Mirage-F1 (BE/CE/CT/EE/M-CE/M-EE/C-200) and F-5E — they can only carry laser bombs and have no self-designation pod, so they cannot employ the weapon autonomously.

Every loadout was validated against pydcs pylon data (mods via their sibling presets).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # Retribution v1.6.0
 
 ## Features/Improvements
+* **[Loadouts]** Aircraft where DEAD isn't a natural role (A-10, F-35, Su-57, Su-35S, strategic bombers, etc.) can now fly DEAD when assigned manually, with a proper standoff loadout, without DEAD being auto-assigned to them by default. Aircraft yamls gain an optional `secondary_tasks` list for this.
 * **[UX]** Hovering a friendly flight's route line on the map highlights it in yellow, and clicking it selects that flight's package (and the flight) in the ATO sidebar.
 * **[UX]** Press Delete with a package selected in the Packages list to cancel it, making it quick to clear several packages in a row.
 * **[UX]** Avoid having escorts from wondering off too far while chasing a target.
--- a/game/campaignloader/squadrondefgenerator.py
+++ b/game/campaignloader/squadrondefgenerator.py
@@ -50,7 +50,8 @@ class SquadronDefGenerator:
             aircraft=aircraft,
             livery=None,
             livery_set=[],
-            auto_assignable_mission_types=set(aircraft.iter_task_capabilities()),
+            auto_assignable_mission_types=set(aircraft.iter_task_capabilities())
+            - aircraft.secondary_tasks,
             radio_presets={},
             operating_bases=OperatingBases.default_for_aircraft(aircraft),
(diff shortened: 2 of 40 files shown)


Starfire13:
But if you're tasking them manually, you would already have access to BAI, which is identical to DEAD, or you could use Strike.

juanjux:
> But if you're tasking them manually, you would already have access to BAI, which is identical to DEAD, or you could use Strike.

Two reasons:
1. Target type: BAI generally is for less dangerous units like tanks and generally uses better weapons for moving targets (but less range) like LGBs or Mavericks. Strike usually is for big stationary targets like buildings and the payloads reflect it generally using big (2000 pound) JDAM or LGB bombs. I think that even while right now many of the weapons for these tasks overlap (because of lack of choice) having DEAD with standoff weapons for dangerous targets adds over these two mission types and allow us to have better payloads for the tasks in the future if the mods improve the range of available weapons.

3. In-game AI tasking (and thus behaviour) is different: DEAD use SEAD inside the game, BAI uses CAS and Strike uses ground attack. This affects how they handle target waypoints, what they do on winchester and many other things.

juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/4

PR #819, opened 2026-06-20
feat(tracker): show runway destruction in mid-mission progress tracker

### LINKED ISSUE #285: Feature Request List

As requested, I am compiling a list of the various features I'd like to see in Retribution.

- [x] **1. Make CAS mission plannable over control points.**

With air assaults now being such a good way to capture enemy airfields and FOBs, it would be nice to have some way to provide CAS for the troops that are dropped off. I've been planning such missions over the frontline and manually dragging the waypoints over the control point for a while now, and while that works fine, this isn't an option on the campaigns that have no frontline.

I'm finding control point CAS to be very handy because more often than not, there's some pesky single half dead ground unit somewhere that the BAI flights didn't manage to demolish, preventing capture. Having a CAS flight swing by and clear out those few remaining ground units for the air assault troops is super helpful.

- [ ] **2. Definable campaign-specific win/loss conditions**

It would be good if we had win/loss conditions that extended beyond just capturing every control point. I think this would help diversify the types of campaign scenarios that are possible. This could be defined in the campaign yaml. As an example, the ability to win a campaign by destroying a specific map object strike target. This could be something like a vital bridge heavily defended by AAA.

- [ ] **3. Campaign win pop up with stats**

When a campaign is won, it would be good if we got a pop up showing the count for all the various units killed by type. Kinda like the end of turn screen, but across the entire campaign. And with stats for both OPFOR and OWNFOR. We could even include the air and ground unit count graphs that are already available in Retribution.

- [x] **4. Campaign options settings autoload from Default.zip**

At the moment, we have to click "load settings", select the campaign options settings zip file, and click open each time we start a campaign. It would be nice if we could create a default settings file (e.g. Default.zip) and that just autoloads as the default. I pretty much always use the same settings.

- [x] **5. Have some way to set "Squadrons start at full capacity" to default on**

I'd like some way to have this set to default on so I don't have to keep remembering to click it each time. Would be good to move it under the Gameplay section of the Mission Generator settings so you can load it from the Campaign Options settings file.

- [ ] **6. Plannable DEAD mission against static armour groups that have a secondary air defence group**

Some static armour ground unit groups have extra tough air defence (e.g. SA-15 TOR). But as it is, a BAI flight will always attack the primary group first and ignore the secondary group until the primary group is dead. This forces us to use longer range stand-off BAI loadouts for the AI so they don't all get shot down. SEAD Sweep is one way to reduce the impact of this problem, but it is not a perfect solution because it is a search and destroy in area task rather than a targeted one. It would be good if there were a way to plan a DEAD mission that targets the secondary group directly for static armour groups, so we could clear out the group's air defence before bombing them.

- [ ] **7. More intelligent grouping of frontline units**

I realise the frontline ground units have the intelligence of a baked potato, and perhaps trying to keep them alive is a lost cause. However, maybe we can help them die in a prettier manner. I'd like to suggest that rather than deploying ground units by order of purchase, to pick a selection from the pool and to then deploy them in mixed groups. This way, if you buy 10 Abrams and 10 Linebackers, you won't always have 10 Abrams on and 5 Linebackers on the 15-unit frontline, and 5 Linebackers in reserve. What's more, the Linebackers won't all be clustered to one side of the frontline. Rather, a mix of 15 Abrams and M113s will be spread evenly across the frontline. This is particularly useful for air defence units, so that they're actually embedded with different armour groups along the frontline so they can defend other units against CA flights instead of being all clustered together.

- [x] **8. Loadout selection drop down menu in Create Flight**

At the moment, changing the loadout for a flight requires creating the flight, then entering the Edit Flight pane, selecting the Payload tab, and changing it in there. What would be a nice QOL feature is if we could select the loadout for a flight right on the Create Flight pane. This is very handy because I often have to switch to alternate loadout presets for AI flights. For example, I want some SEAD flights to carry HARMs, others to carry TALDs. Or I might want certain BAI flights to use mavericks instead of cluster bombs if they are attacking static armour group that have SHORAD.

- [ ] **9. Show runway destruction in mid-mission progress tracker**

At the moment, the mid-mission progress tracker tracks unit deaths, base captures, etc. But it doesn't track runway destruction. Would be nice if runway destruction also shows up there.

- [ ] **10. Air Wing Configuration faction tabs should be renamed to OWNFOR and OPFOR**

At the moment, they are labelled "Blue" and "Red" which would not always be accurate since you can play as Redfor in Retribution. Heck, you can even play Red vs Red if you wanted to, or Blue vs Green.

Solution I'd like is for Stoner to win a billion dollars in the lottery so he can work on Retribution full time.

### Additional context

_No response_

geofffranks:
## What

  The mid-mission progress tracker already received destroyed-airfield data
  via `Debriefing.damaged_runways`, but never surfaced it to the player. This
  adds a **"Runways destroyed"** count to the tracker grid so runway kills show
  up alongside the other mid-mission loss/progress stats.

  ## How

  - Add `runways_destroyed` to `SideLossCounts` / `loss_counts` in
    `game/debriefing.py`, derived from the existing `damaged_runways` data.
  - Add a "Runways destroyed" row to the tracker grid in
    `QWaitingForMissionResultWindow.py`.
  - Cover the new count in `tests/test_debriefing.py`.

  No new data plumbing — the debrief already carried the runway info; this just
  counts and displays it.

  Closes #285 (item #9).

### DIFF
--- a/game/debriefing.py
+++ b/game/debriefing.py
@@ -102,6 +102,7 @@ class SideLossCounts:
     ground_objects: int
     scenery: int
     bases_lost: int
+    runways_destroyed: int
 
 
 @dataclass(frozen=True)
@@ -305,6 +306,7 @@ class Debriefing:
             airlifts = gl.player_airlifts
             ground_objects = gl.player_ground_objects
             scenery = gl.player_scenery
+            airfields = gl.player_airfields
         else:
             air = self.air_losses.enemy
             front_line = gl.enemy_front_line
@@ -313,6 +315,7 @@ class Debriefing:
             airlifts = gl.enemy_airlifts
             ground_objects = gl.enemy_ground_objects
             scenery = gl.enemy_scenery
+            airfields = gl.enemy_airfields
         return SideLossCounts(
             aircraft=len(air),
             front_line=len(front_line),
@@ -326,6 +329,7 @@ class Debriefing:
                 for capture in self.base_captures
                 if capture.captured_by_player == player.opponent
             ),
+            runways_destroyed=len(airfields),
         )
 
     def dead_aircraft(self) -> AirLosses:
--- a/qt_ui/windows/QWaitingForMissionResultWindow.py
+++ b/qt_ui/windows/QWaitingForMissionResultWindow.py
@@ -172,6 +172,7 @@ class QWaitingForMissionResultWindow(QDialog):
             ("Ground Objects lost", blue.ground_objects, red.ground_objects),
             ("Scenery Objects lost", blue.scenery, red.scenery),
             ("Bases lost", blue.bases_lost, red.bases_lost),
+            ("Runways destroyed", blue.runways_destroyed, red.runways_destroyed),
         ]
         for label, blue_count, red_count in rows:
             self.add_update_row(label, blue_count, red_count, update_layout)
--- a/tests/test_debriefing.py
+++ b/tests/test_debriefing.py
@@ -44,6 +44,8 @@ def _sample_debriefing() -> Debriefing:
         enemy_ground_objects=_items(2),
         player_scenery=_items(0),
         enemy_scenery=_items(1),
+        player_airfields=_items(1),
+        enemy_airfields=_items(2),
     )
     captures = [
         _capture(Player.BLUE),
@@ -68,6 +70,7 @@ def test_loss_counts_blue_side() -> None:
         ground_objects=5,
         scenery=0,
         bases_lost=1,  # one base captured by RED == one base Blue lost
+        runways_destroyed=1,
     )
 
 
@@ -82,6 +85,7 @@ def test_loss_counts_red_side() -> None:
         ground_objects=2,
         scenery=1,
         bases_lost=2,  # two bases captured by BLUE == two bases Red lost
+        runways_destroyed=2,
     )
 
 
@@ -105,3 +109,6 @@ def test_loss_counts_partition_matches_combined_totals() -> None:
     )
     assert blue.scenery + red.scenery == len(list(debriefing.scenery_object_losses))
     assert blue.bases_lost + red.bases_lost == len(debriefing.base_captures)
+    assert blue.runways_destroyed + red.runways_destroyed == len(
+        list(debriefing.damaged_runways)
+    )


Starfire13:
My recollection is that there are multiple events logged when multiple bombs are dropped on the same runway. Does this know how to tell they are duplicates?

geofffranks:
I'll have to give that another try. I validated bombing on blue and red runways, and at least for me, I needed to drop about 8,000 lbs worth of payload to get the runway to be marked as destroyed in the first place. i've never tried 16k to see if it thinks 2 are destroyed, but i'll get a campaign/misssion with closer airfields and try 4 round trips

Starfire13:
You should be able to get a runway kill event with a single 2000lb JDAM. GBU-31v3 is what I use, though any variant of the GBU-31 will work. So if you drop, say, 4 GBU-31s, you will get 4 kill events in the log.

geofffranks:
I've never been able to get a single gbu-31v3 penetrator to count as a runway destruction (at least in an F-16). It's always had to be a lot more. I just ran another run, dropped 2 of those, and then 2 paveway 2k bombs.  It finally scored a destruction. I ran a few more runs and dropped a total of 8-10k lbs on the runway and only saw one destruction recorded.

Starfire13:
Odd. I've been using single GBU-31s to kill runways for years, both by myself as well as with AI. It has never failed. Do you have splash damage plugin enabled?

geofffranks:
We typically do but I should double check the campaign I was testing on.Sent from my iPhoneOn Jun 23, 2026, at 1:19 AM, Starfire13 ***@***.***> wrote:﻿Starfire13 left a comment (dcs-retribution/dcs-retribution#819)
Odd. I've been using single GBU-31s to kill runways for years, both by myself as well as with AI. It has never failed. Do you have splash damage plugin enabled?

geofffranks:
Yeah, that was it. Apparently we've been running without splash damage for at least the last campaign. No wonder things have been miserable. 

Turned it on, and one GBU-31 triggered runway destruction. Dropped a second, and two paveways for good measure, still 1 runway destroyed.

Druss99 (approved):

PR #820, opened 2026-06-20
feat(waypoint): propagate edited waypoint names to the aircraft CDU/HUD

### LINKED ISSUE #695: Allow Edited Waypoint Names to Appear in Aircraft Flight Plan

Currently, if I utilize the "Add Waypoint" feature to add a set of target locations, they display with a long form name such as: "[ARAPAIMA] : SSM SS-1C Scud-B #0" , "[ARAPAIMA] : SSM SS-1C Scud-B #1" , etc. I can edit the names in the Retribution flight plan to be more succinct such as: SCUD1, SCUD2. However, the edited name doesn't appear in the A-10C flight plan and instead the long form name still displays. These waypoint names are too long to display properly in the HUD or CDU flight plan and it's difficult to tell which target waypoint is selected. It appears that if I open the mission in the DCS Mission Editor, I can modify the waypoint names there, but it is somewhat time consuming to do so.  

As suggested in the Discord channel, I'm requesting an enhancement that allows waypoint name edits from the Retribution flight plan editor to be injected into the mission file.

### Additional context

_No response_

geofffranks:
## What

  When you rename a waypoint in the flight plan editor, that custom name now
  follows through to the aircraft's CDU/HUD in-cockpit — not just the
  Retribution UI list and kneeboard. Previously the cockpit always showed the
  long auto-generated name regardless of your edit.
  One name, three places: flight plan list, kneeboard, and the in-cockpit CDU.

  ## How

  - `FlightWaypoint` gains a `custom_name` override (`game/ato/flightwaypoint.py`):
    - `display_name` returns `custom_name or pretty_name`.
    - `apply_name_edit(raw)` stores the trimmed name, or `None` when it's empty
      or equals the auto name (so it never persists a phantom override).
  - `dcs_name_for_waypoint()` (`pydcswaypointbuilder.py`) emits `custom_name`
    into the `.miz` for normal waypoints, with the F-15E target page wired the
    same way.
  - Structural routing markers are left untouched — `JOIN`, `SPLIT`,
    `RACETRACK START`, `RACETRACK END`, `DROPOFFZONE` keep their canonical
    cockpit names even if renamed in the list (rename is a harmless no-op for
    those).
  - UI list + kneeboard wiring so the edited name shows everywhere.

  ## Scope / notes

  - Renamed-only: a waypoint left at its default auto name behaves exactly as
    before; only an explicit edit propagates.
  - AI-only flights are unaffected — target/divert/bullseye waypoints are
    player-only and never emitted for AI flights, so there's nothing to
    propagate there.

  Closes #695.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -20,6 +20,7 @@
 * **[Modding]** Update to CJS Super Hornet Mod to v2.4.5.260501.RC1
 * **[Modding]** Update Community A4EC Mod to 2.3.0 (May 2025)
 
+* **[Flight Plans]** Renaming a waypoint in the flight-plan list now propagates to the aircraft CDU/HUD and the kneeboard, not just the list — one name in all three places (#695).
 
 ## Fixes
 * **[Plugins]** Fix the escort leash never running (DCS has no `Group.getByID`; look the group up by name via mist), so escorts are actually held to their engagement range.
--- a/game/ato/flightwaypoint.py
+++ b/game/ato/flightwaypoint.py
@@ -34,6 +34,9 @@ class FlightWaypoint:
     targets: list[TheaterUnit] = field(default_factory=list)
     obj_name: str = ""
     pretty_name: str = ""
+    # User override; None = follow the auto names. Invariant: never "" — apply_name_edit
+    # stores None instead of an empty string, so `custom_name or pretty_name` is safe.
+    custom_name: str | None = None
     only_for_player: bool = False
     flyover: bool = False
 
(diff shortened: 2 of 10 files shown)


Druss99 (approved):

Druss99:
Before merging can we update the change log and also I don't see anything regarding excluding this for AI though the description mentions it is.

geofffranks:
> Before merging can we update the change log and also I don't see anything regarding excluding this for AI though the description mentions it is.

Changelog updated. I added some comments/tests related to validating it excludes AI flights, but the gist is:

> On the AI exclusion: it's structural rather than an explicit guard in the name resolver. `dcs_name_for_waypoint()` reads only the waypoint's own `custom_name`, and a rename is written only to the player's own flight-plan waypoints (`FlightWaypoint.apply_name_edit`), so an AI flight's waypoints never carry one. On top of that, the waypoint types a rename usually targets — target, divert, bullseye — are built `only_for_player=True` and are filtered out for AI flights (`client_count == 0`) in `WaypointGenerator.create_waypoints`.

PR #822, opened 2026-06-21
fix(ctld): treat landed helo as on-ground via terrain-AGL

### LINKED ISSUE #725: Too high or too fast to drop troops/cargo when landed or within acceptable constraits bug
### Affected versions

v1.4.1

### Test/Development build

Build #2461

### Description

So this is the second time I have carried out a pre-generated air assault mission flying the Ch-47F. I fly out to the island to be assaulted, once there I land to disembark troops using CTLD. I get a notice that I am flying to high or too fast yet I am landed and stationary. I also tried hovering at 10ft and 0 speed and I still cannot complete the mission. This was on the Scenic Route Merged Campaign in Retribution. I was not able to complete a mission on mainland Iran at Bandar Bengeh or at Tunb Island because of this bug. I wish there was some way to force troops or cargo to disembark when this bug happens but there is no workaround. 

### Save game and other files

[retribution_nextturn.zip](https://github.com/user-attachments/files/26649305/retribution_nextturn.zip)

geofffranks:
CTLD blocked troop/cargo disembark with "Too high or too fast" even when a CH-47F was landed and stationary. The CH-47F (a newer DCS module) reports unit:inAir() == true while on the ground, so ctld.inAir() fell through to safeToFastRope(), which then failed on ground-vibration velocity readings.

Override ctld.inAir in the Retribution tailor-CTLD config to trust terrain-AGL as ground truth: within 1 m of the ground counts as landed, regardless of the unreliable :inAir() call. Placed in ctld-config.lua (not the vendored CTLD.lua) so it survives wholesale CTLD re-vendoring; it loads after CTLD.lua and redefines the global.

Fixes #725.

### DIFF
--- a/resources/plugins/ctld/ctld-config.lua
+++ b/resources/plugins/ctld/ctld-config.lua
@@ -88,6 +88,29 @@ if dcsRetribution then
                 end
             end
 
+            --- CH-47F (and some other DCS modules) report :inAir() == true while
+            --- on the ground, so CTLD wrongly treats a landed helicopter as
+            --- airborne and blocks troop/cargo unload with "too high or too fast".
+            --- Override ctld.inAir to trust terrain-AGL as ground truth: within
+            --- 1 m of the ground counts as landed (a margin above 0 so a model
+            --- whose origin sits slightly above its contact points, or a unit on
+            --- a gentle slope, still reads as landed). See dcs-retribution #725.
+            --- Intentionally outside the tailorctld block: the :inAir() bug is
+            --- config-independent, so this override applies to every mission.
+            function ctld.inAir(_heli)
+                if _heli:inAir() == false then
+                    return false
+                end
+                if ctld.heightDiff(_heli) < 1.0 then
+                    return false
+                end
+                -- less than 5 cm/s so landed, BUT AI can hold a perfect hover so ignore AI
+                if mist.vec.mag(_heli:getVelocity()) < 0.05 and _heli:getPlayerName() ~= nil then
+                    return false
+                end
+                return true
+            end
+
             --- add all carriers as pickup zone
             if dcsRetribution.Carriers then
                 for _, carrier in pairs(dcsRetribution.Carriers) do


resources/plugins/ctld/ctld-config.lua
```diff
@@ -88,6 +88,29 @@ if dcsRetribution then
...
+            --- 1 m of the ground counts as landed (a margin above 0 so a model
+            --- whose origin sits slightly above its contact points, or a unit on
+            --- a gentle slope, still reads as landed). See dcs-retribution #725.
+            --- Intentionally outside the tailorctld block: the :inAir() bug is
+            --- config-independent, so this override applies to every mission.
+            function ctld.inAir(_heli)
+                if _heli:inAir() == false then
+                    return false
+                end
+                if ctld.heightDiff(_heli) < 1.0 then
```
Druss99, line 104: I think we should combine this with an and statement for speed to ensure you're relatively stable and likely on the ground. I also think the AI check isn't needed since they'll just drop troops whenever they meet the conditions anyway

Druss99 (changes requested):

resources/plugins/ctld/ctld-config.lua
geofffranks: The AI check was there in the upstream  ctld code, but I'm fine removing it. 

I'm worried that adding the velocity check might bring back the original problem though. The too high or too fast message was generated after [failing this check](https://github.com/dcs-retribution/dcs-retribution/blob/b5c3e0a38162d5ac451f06b19f1611dc9a14e786/resources/plugins/ctld/CTLD.lua#L1855) to see if the helicopter is either not [ctld:inAir()](https://github.com/dcs-retribution/dcs-retribution/blob/b5c3e0a38162d5ac451f06b19f1611dc9a14e786/resources/plugins/ctld/CTLD.lua#L1834-L1846) or [ctld:safeToFastRope()](https://github.com/geofffranks/dcs-retribution/blob/b5c3e0a38162d5ac451f06b19f1611dc9a14e786/resources/plugins/ctld/CTLD.lua#L1815-L1825). 
  
So both ctld:inAir() and ctld:safeToFastRope() failed. If safeToFastRope() failed, it means that we have enableFastRopeInsertion set to false (which afaict it's always true in ctld?), and either the speed was >2.2m/s, or height was >21.28m. Since the above fix used the same heightDiff function and it was <1, it means the velocity of the landed CH-47 was > 2.2m/s. 🤯

So I *think* there's probably a bug in both :inAir() and :getVelocity() on the CH-47 implementation, or the helicopter was sliding at about 4kts while on the ground.

Either way making the velocity check an `and` here would make the bug come back, right?

Druss99 (approved):

PR #824, opened 2026-06-22
Plugins: support drop-down (choice) options in plugin settings

juanjux:
Plugin `specificOptions` could only be booleans (checkbox) or numbers (spin box). This adds a third type: an option that declares a `"choices"` list of strings renders as a **drop-down (`QComboBox`)** in the plugin settings.

- `game/plugins/luaplugin.py`: `LuaPluginOption` now carries `choices`, and `from_json` reads the optional `"choices"` key. Plugin-option values are injected into the mission Lua via a small `_lua_literal` helper that **quotes strings** — previously a non-bool/number value was written unquoted (`... = foo`), which Lua reads as an undefined identifier. `bool` (`true`/`false`) and number injection are unchanged.
- `qt_ui/windows/settings/plugins.py`: renders a `QComboBox` for choice options and restores its value in `update_from_settings`.

No behaviour change for existing plugins (none declare `choices`). `black` clean.

Example option:
```json
{ "nameInUI": "Message audience", "mnemonic": "SCOPE", "choices": ["A", "B", "C"], "defaultValue": "B" }
```

(Note: This is used in a branch I am testing and will publish soon using the EW script for more features).

### DIFF
--- a/game/plugins/luaplugin.py
+++ b/game/plugins/luaplugin.py
@@ -58,10 +58,19 @@ class PluginSettings:
 
 class LuaPluginOption(PluginSettings):
     def __init__(
-        self, identifier: str, name: str, min: Any, max: Any, value: Any
+        self,
+        identifier: str,
+        name: str,
+        min: Any,
+        max: Any,
+        value: Any,
+        choices: Optional[List[str]] = None,
     ) -> None:
         super().__init__(identifier, value)
         self.name = name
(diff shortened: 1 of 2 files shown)


juanjux:
Closing — consolidating ongoing development in my fork to ease the upstream review load. This PR continues here (also easy for other forks to pick up): https://github.com/juanjux/dcs-retribution/pull/2

PR #827, opened 2026-06-23
Campaign doctrine: ignore non-combat (crash) air losses

juanjux:
Adds a **Campaign Doctrine → General** option *"Non-combat (crash) aircraft losses don't count"* (default off). When enabled, air losses DCS does not credit to a weapon or SAM — AI crashes, collisions, terrain/fuel — no longer deplete the squadron and the pilot survives; only real combat kills count. Applies to both coalitions.

Implemented on top of new per-loss **kill attribution** in `Debriefing`: each `S_EVENT_KILL` is indexed by killed unit and resolved to its loss object (including TIC-respawned front-line clones), so a loss can be classified as non-combat by elimination. The debriefing window now shows, per aircraft type, how many losses were not counted.


juanjux:
Closing — I'm consolidating ongoing development in my fork to ease the upstream review load. This PR continues here, where it can also be picked up by other forks: https://github.com/juanjux/dcs-retribution/pull/1

PR #833, opened 2026-06-24
Mission Planning: carrier/LHA targets offer SEAD instead of a duplicate SEAD Escort

juanjux:
Carrier and LHA targets listed **SEAD Escort twice** in the "Create flight" task menu: `NavalControlPoint.mission_types()` yielded `SEAD_ESCORT` explicitly, but the base `MissionTarget.mission_types()` already yields it. This changes the explicit yield to **SEAD**, so carrier groups offer SEAD (their escorts are SAM platforms and can be suppressed directly, like any other naval group) with no duplicate — `SEAD_ESCORT` is still available via the base class.

A new parametrized test (`tests/theater/test_mission_types_no_duplicates.py`) asserts that no concrete `MissionTarget` offers a duplicate `FlightType`, for either coalition.

Verified: the new test passes (40 cases); reverting only the one-line fix makes it fail for exactly `Carrier`, `EssexCarrier` and `Lha`; the full `tests/theater` suite passes with no regressions.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -19,6 +19,7 @@
 * **[Options]** Add new option to fast forward until player is at the IP.
 
 ## Fixes
+* **[Mission Planning]** Carrier/LHA targets now offer SEAD in the flight-task list and no longer list SEAD Escort twice (their escorts are SAM platforms, so they can be suppressed directly like any other naval group).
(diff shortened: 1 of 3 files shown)


Druss99 (approved):

Druss99:
Ideally we would guard against planning SEAD if only the carrier is alive but I think that can be a follow up

PR #838, opened 2026-06-25
Air assault: keep the drop-off zone out of the water

juanjux:
For coastal objectives the air-assault drop-off position was computed without a land check: the CTLD random point (`random_point_within`) or the 1200 m offset toward the ingress could land offshore, dropping the troops into the sea (where they die, DCS soldiers doesn't swim very well).

Snap the drop-off to the nearest land with `theater.nearest_land_pos` — a no-op when the point is already on land, and on theaters that ship no landmap.

### DIFF
--- a/game/ato/flightplans/airassault.py
+++ b/game/ato/flightplans/airassault.py
@@ -154,9 +154,7 @@ class Builder(FormationAttackBuilder[AirAssaultFlightPlan, AirAssaultLayout]):
         else:
             heading = tgt.position.heading_between_point(ingress.position)
             drop_pos = tgt.position.point_from_heading(heading, 1200)
-            # Keep the drop-off out of the water: for coastal objectives the
-            # 1200 m offset toward the ingress can land offshore. Predefined CTLD
-            # drop-off zones are placed by the campaign designer, so leave them as-is.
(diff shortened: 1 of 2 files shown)


game/ato/flightplans/airassault.py
```diff
@@ -154,6 +154,8 @@ def layout(self) -> AirAssaultLayout:
         else:
             heading = tgt.position.heading_between_point(ingress.position)
             drop_pos = tgt.position.point_from_heading(heading, 1200)
+
+        drop_pos = self.flight.coalition.game.theater.nearest_land_pos(drop_pos)
```
Druss99: This seems to be ignoring the predefined CTLD drop off points above. On mobile so I can't confirm but I think this should go under the else statement

Druss99 (changes requested):

game/ato/flightplans/airassault.py
juanjux: Good catch! You are right, it should go in the `else`. This should be fixed now.

juanjux: Change done and conflict fixed, PTAL when you can @Druss99

Druss99 (approved):

PR #843, opened 2026-06-27
Era-gate payload-editor options: JHMCS helmet cueing + targeting pods

BradySox:
## What

Extends the existing **`restrict_weapons_by_date`** toggle from weapons to two more payload-editor choices, so a pre-era campaign stops fielding anachronistic kit. No-op unless that setting is on.

### 1. JHMCS helmet cueing — aircraft *properties*

DCS unit "properties" (the payload-editor mission options) carry no introduction year, so this adds a small curated date table:

- **New `game/dcs/aircraftproperties.py`** — keyed by the property value **label**, not the numeric id. That's deliberate: `HelmetMountedDevice` id `1` is **"JHMCS"** on the F/A-18 and F-16 (fielded ~2003) but **"SURA Visor"** (a 1980s Soviet sight) on the Su-30/Su-35 — an id-based gate would wrongly restrict the Soviet sight. The gate is further scoped to the helmet-device identifiers.
- **`FlightGroupConfigurator.degrade_props_for_date`** is authoritative: it clamps gated properties to a period-correct value at generation, resolving against the unit-type default so the common *defaulted*-JHMCS case is caught (an unset helmet still defaults to JHMCS in the `.miz`).
- The payload-editor dropdown filters gated options out for the campaign date (UI parity).

### 2. Targeting pods — weapon era data (restores withdrawn #786)

Adds fielding-year + fallback data for **AN/AAQ-33, DAMOCLES, AN/AAS-38 Nite Hawk, AN/AVQ-23 Pave Spike, AN/AVQ-26 Pave Tack, AVIC WMD-7**, plus the missing **ATFLIR/LITENING** CLSID variants — so older airframes degrade to a period-correct pod instead of falling back to "unknown". (This re-opens the previously-withdrawn #786 on current `dev`.)

## Tests

- `tests/dcs/test_aircraftproperties.py` — JHMCS gating, baseline/NVG always-available, SURA-Visor-not-gated, non-helmet-property-untouched.
- `tests/data/test_weapons.py` — pod era data + a guard that every targeting pod referenced in `customized_payloads` resolves to dated TGP data.

## Compatibility

Fully backward-compatible and behavior-preserving when `restrict_weapons_by_date` is off (the default). No save-format change.


Druss99 (changes requested):
I think we need to base this around how weapons are done where we have YAML files for the different helmets and their introduction year

BradySox:
> I think we need to base this around how weapons are done where we have YAML files for the different helmets and their introduction year

Makes sense — moved the helmet-cueing era data into YAML mirroring the weapons model (resources/aircraftproperties/helmets/*.yaml, one entry per system with name / introduction_year / property_ids), loaded at import; the gating functions and both consumers are unchanged. While doing it I also closed a gap: id 1 of HelmetMountedDevice is a different era-defining sight on each airframe, so I gated each at its own year rather than only JHMCS — JHMCS (2003), the MiG-29 HMS / Shchel-3UM (1985), the Su-30 SURA Visor (1996), and the A-10C II HMCS / “HMCS + NVG” Scorpion (2012). That’s exactly why the data is keyed by the value label not the numeric id (an id gate would force one year on all of them). Plain NVG and the baseline options stay available in every era.

game/dcs/aircraftproperties.py
```diff
@@ -0,0 +1,130 @@
...
+    if not introduction_years:
+        logging.warning("No helmet-cueing date data found in %s", _HELMET_DATA_DIR)
+    return frozenset(property_ids), introduction_years
+
+
+#: Property identifiers that carry a helmet-mounted cueing selection. Scoping the gate to
+#: these keeps it from touching any unrelated property that shares a gated value label.
+#: Helmet-mounted cueing systems gated by their real-world fielding year, keyed by the
+#: pydcs value label (loaded from ``resources/aircraftproperties/helmets/*.yaml``).
+HELMET_DEVICE_PROPERTY_IDS, HELMET_CUEING_INTRODUCTION_YEARS = (
```
Druss99: Lets use a class in this file so we don't need to make a bunch of globals

game/missiongenerator/aircraft/flightgroupconfigurator.py
```diff
@@ -353,9 +354,32 @@ def setup_props(self) -> None:
                     props.update(laser_code_config.property_dict_for_code(code.code))
             if unit.unit_type.datalink_networkable() and self.no_datalink_set(props):
                 self.set_datalink(props, unit.callsign_as_str())
+            if self.game.settings.restrict_weapons_by_date:
```
Druss99: I think we probably want a new setting for this so that users can choose to enforce weapons, HMCS, or both

resources/aircraftproperties/helmets/HMCS-NVG.yaml
```diff
@@ -0,0 +1,7 @@
+# The A-10C II "HMCS + NVG" option pairs the Scorpion HMCS (see HMCS.yaml) with NVGs.
+# Because it carries the era-defining HMCS, it is gated at the same year. Plain NVGs are
+# not gated; only the HMCS-bearing options are.
+name: HMCS + NVG
```
Druss99: Maybe instead of separate files these should live in the aircraft files since they are aircraft specific? Alternatively they could live here but they should probably have a field for the aircraft or the file should be named after the aircraft that way we don't have to iterate over every aircraft to check if it has one of these helmets as that is a lot of iterations

resources/weapons/pods/litening.yaml
```diff
@@ -13,3 +13,9 @@ fallback: AIM-120C
 clsids:
   - "{A111396E-D3E8-4b9c-8AC9-2432489304D5}"
   - "{AAQ-28_LEFT}"
+  - "{F-15E_AAQ-28_LITENING}"
+  - "{HB_LITENING_ON_ADAPTER_IN_AERO7}"
```
Druss99, line 17: Lets pull these weapon updates into a separate PR

Druss99 (changes requested):

game/dcs/aircraftproperties.py
BradySox: Done — the module is now a single frozen `PropertyDateGate` dataclass; the module-level globals (and the loader that populated them) are gone.

game/missiongenerator/aircraft/flightgroupconfigurator.py
BradySox: Done — added an independent `restrict_props_by_date` setting (next to the weapons toggle in Campaign Management); the generator clamp and the payload-editor dropdown now gate on it, so weapons and aircraft options can be enforced separately or together.

resources/aircraftproperties/helmets/HMCS-NVG.yaml
BradySox: Done — the era data moved into each aircraft's own `resources/units/aircraft/<type>.yaml` as a `date_gated_properties` block, loading into `AircraftType.property_date_gate` at unit-data load. Lookup is a direct attribute access on the airframe being configured — no iteration over other aircraft, and the `resources/aircraftproperties/` directory is gone. One note from the move: SURA Visor was dropped — no pydcs airframe exposes that value label (the Su-30 is a mod), so it never gated anything; it comes back as a two-line data edit in the mod's yaml if its label is confirmed. A new test pins every gated label against pydcs so a rename can't silently kill a gate.

resources/weapons/pods/litening.yaml
BradySox: Done — the pod updates are reverted out of this branch and opened as #871.

qt_ui/windows/mission/flight/payload/propertycombobox.py
```diff
@@ -1,17 +1,25 @@
 from PySide6.QtWidgets import QComboBox
 from dcs.unitpropertydescription import UnitPropertyDescription
 
+from game import Game
```
Druss99: If this is only needed for type checking lets gate it on TYPE_CHECKING so we don't need to initialize the game object

Druss99 (changes requested):

Druss99:
Also please update the changelog

Druss99 (approved):

PR #844, opened 2026-06-28
Removes unused mission_types field from squadron definitions

geofffranks:
Found this while cleaning up the wiki. It turns out we don't load mission_types from squadron definitions ever. It's all based on the aircraft yaml.

Flow:

1. Base capability = aircraft type, fixed. capable_of(t) / iter_task_capabilities() come from AircraftType. Squadron can't grant itself a task the airframe can't do.
2. Def default (squadrondef.py:110): auto_assignable_mission_types = set(unit_type.iter_task_capabilities()) → starts as all aircraft caps. (The dead squadron-YAML mission_types key never fed this — now removed.)
3. Filter, not redefine (squadron.py:173-178): def set_auto_assignable_mission_types(self, mission_types):
    self.auto_assignable_mission_types = {t for t in mission_types if self.capable_of(t)}
3. Intersects request with capable_of. Can only shrink the auto-plannable set, never add capability.
4. External config that drives the override = the campaign air-wing config (defaultsquadronassigner.py:64): squadron.set_auto_assignable_mission_types(squadron_config.auto_assignable)
4. where (campaignairwingconfig.py:31): auto_assignable = set(self.secondary) | {self.primary}
4. So the campaign yaml's per-base primary: + secondary: keys (secondary can be alias any/air-to-air/air-to-ground) define what the squadron auto-plans — then filtered by aircraft caps.
5. migrator.py:200 just re-runs the filter on save reload. No new source.

Bottom line:
- auto_assignable_mission_types = "which of my airframe's capabilities the auto-planner may use" — a subset, configured by campaign primary/secondary, never by squadron def YAML.
- Squadron capabilities themselves = aircraft type, immutable.
- Removing the squadron-YAML mission_types key changed nothing — it was never wired to this path. Safe.


Druss99:
I had to do a double take but this is actually true though I think the answer won't be stripping the mission types out of squadron files but instead parsing them correctly.

That being said it might have some interesting effects on randomized squadrons if we don't check that the squadron is able to perform the primary mission type specified in the campaign yaml

geofffranks:
Closing in favor of https://github.com/dcs-retribution/dcs-retribution/pull/849

PR #847, opened 2026-06-29
Improve F-4E-45MC loadouts and fix Maverick date-fallback

BradySox:
Two related fixes for the Heatblur F-4E (and Mavericks generally).

## 1. F-4E-45MC default loadouts

The `F-4E-45MC` Retribution presets were a grab-bag of odd single-weapon fits that read as nonsensical in the payload editor. Rebuild all 13 from the module's own built-in loadouts, mapped to the matching `Retribution <task>` names, with a period-correct **AIM-7E2 / AIM-9L** air-to-air baseline (AIM-7F/9P on BARCAP) instead of the previous all-modern AIM-7M — so `restrict_weapons_by_date` degrades to best-available across both early and modern eras.

## 2. Maverick date-fallback → Rockeye, not Walleye

With date restriction on, every AGM-65 Maverick's fallback chain bottomed out at the **AGM-62 Walleye** (`AGM-65D → AGM-65B → AGM-65A → AGM-62 Walleye`). A Maverick is a CAS/anti-armor PGM; degrading it to a ~1100 lb TV-guided standoff bomb is the wrong mission — and it made the F-4E's CAS/DEAD presets carry Walleyes on pre-1972 campaigns. Reroute `AGM-65A`'s fallback to **Mk-20 Rockeye** (anti-armor cluster, 1968); the whole AGM-65 family chains through that node, so it fixes every aircraft's Maverick degrade:

> `AGM-65D → AGM-65B → AGM-65A → Mk-20 Rockeye → Mk 82`

## Validation

Headless against this tree: all 13 F-4E presets resolve, are station-legal, and map to their tasks (no clean-jet fallback); the weapon DB loads and the Maverick chain no longer contains Walleye. Data-only change to two files.

### DIFF
--- a/resources/customized_payloads/F-4E-45MC.lua
+++ b/resources/customized_payloads/F-4E-45MC.lua
@@ -6,84 +6,60 @@ local unitPayloads = {
 			["name"] = "Retribution CAS",
 			["pylons"] = {
 				[1] = {
-					["CLSID"] = "{HB_ALE_40_30_60}",
-					["num"] = 14,
-				},
-				[2] = {
-					["CLSID"] = "{AGM_62_I}",
+					["CLSID"] = "{DB769D48-67D7-42ED-A2BE-108D566C8B1E}",
 					["num"] = 13,
 				},
+				[2] = {
+					["CLSID"] = "{HB_F4E_AIM-7E-2}",
+					["num"] = 9,
+				},
 				[3] = {
-					["CLSID"] = "<CLEAN>",
-					["num"] = 12,
-					["settings"] = {
-						["EAS_bypass_ctrl"] = 1,
-						["NFP_rfgu_type"] = 1,
-						["rf_lower_limit_ctrl_Mk22Mod2"] = 4800000000,
-						["rf_upper_limit_ctrl_Mk22Mod2"] = 5200000000,
-					},
+					["CLSID"] = "{HB_F4E_AIM-7E-2}",
+					["num"] = 8,
 				},
 				[4] = {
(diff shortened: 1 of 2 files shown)


Druss99:
The loadouts are not made to be historically accurate. Loadouts are created based on the best weapons available to make the AI perform well for a mission.

As for the fallbacks I disagree. The walleye should be the fallback since it is a standoff weapon like the maverick. The rockeye is a very different weapon with a very different purpose

Druss99:
Closing as I think these loadouts make the AI less effective. If you'd like you can always create your own custom loadouts to use in place of the defaults

BradySox:
> Closing as I think these loadouts make the AI less effective. If you'd like you can always create your own custom loadouts to use in place of the defaults

Fair enough, happy to keep it local. I was looking specifically at the low bomb totals and unused pylons on some loadouts, when I'm home I'll double check and narrow it down if I still it's needed

Starfire13:
Our intent with the default loadouts has generally been to provide the AI with the best loadout to get the job done rather than strict historical accuracy. Stand-off weapons are preferred where available, guided munitions are preferred where available. The most modern variant of missiles is used where available. We then rely on the restrict weapons by date feature to swap weapons out for period-appropriate ones for those who prefer more historical accuracy.

I don't recall specifically why certain pylons might have no stores when I made the loadouts but in the past, the usual reason is that this would have resulted in mixed loadouts (bombs of different sizes, for example). The AI does not like mixed loadouts on certain mission types.

PR #854, opened 2026-07-03
Spawn each squadron under its own DCS country for nation-specific voiceovers (#627)

### LINKED ISSUE #627: Different countries per side

Retribution campaigns are forced to use just one single country, or CJTF blue/red for all units. Since certain countries on DCS have unique voiceovers for communications, using a single country can lead to non-immersive gameplay due to many units that have liveries of different nations having all the same voices.

Give retribution the ability to select different nations for blue and red sides, allowing for these unique voiceovers to be included in gameplay. Preset squadrons are already tied to unique nations, so maybe it could look like having the faction's country be the default for randomly generated squadrons, but then have retribution divert to whatever nation is called out on the squadron yaml?

### Additional context

_No response_

BradySox:
Resolves #627.

## What

DCS plays nation-specific voiceovers and comms based on the **country** a group belongs to. Retribution collapsed every unit on a side onto a single faction country, so a coalition (CJTF) faction flying squadrons drawn from several nations all shared one nation's radio voice. This spawns each squadron's units under their own DCS country, and names their pilots to match.

The data already existed — squadrons carry a `country` (set by preset YAML `country:`, inherited from the faction by auto-generated squadrons). The gap was purely in mission generation. This routes the squadron country through the spawn path.

## How

- **`CountryAssigner`** (`game/missiongenerator/countryassigner.py`, new) resolves the DCS country per squadron plus the set of countries to register on each coalition. It enforces the one DCS constraint that matters — a country may belong to only **one** coalition in a `.miz`:
  - Blue claims its squadron countries first; a red squadron whose country is already claimed by blue falls back to red's faction country.
  - Each side's faction country stays exclusive to that side (it's the spawn fallback), so a blue squadron that happens to share red's faction country falls back to blue's faction country rather than registering the nation on both coalitions.
  - It interns **one canonical `Country` instance per id** and hands that same instance to both the coalition registration and the spawn call. pydcs attaches groups to the instance via `country.add_aircraft_group` and only serializes countries reachable from the coalition, so a duplicate-id instance would silently drop its groups on save.
- **`MissionGenerator` / `AircraftGenerator`** register all per-side countries and resolve the country **per flight/squadron** in `generate_flights` and `spawn_unused_aircraft` (the old coalition-level `country` params are dropped). `FlightGroupSpawner` is unchanged — it already spawns under whatever country it's handed, and callsign generation flows from the same value.
- **Nation-aware pilot names** complete the arc (`game/squadrons/pilotnames.py`): a curated DCS-country → Faker-locale table so `Squadron.faker` draws from the squadron's own nation (a Greek squadron rosters Greek names, an Iranian one Persian, a Russian one surname-first patronymics). Any unmapped / multinational country (CJTFs, Insurgents, UN Peacekeepers) — or a locale not shipped/gender-aware in the installed Faker — falls back to the coalition faker, so a roster is **always** produced.

## No-op for single-nation factions

For a non-CJTF faction the squadron loader already restricts squadrons to the faction country, so every resolved country equals the faction country and the generated mission is byte-identical. The behavior only diverges for mixed-nation / CJTF sides — exactly the intent.

The Pretense generator is untouched: it overrides `generate_air_units` / `setup_mission_coalitions` and keeps its existing single-country-per-side behavior.

## Tests

- `tests/missiongenerator/test_country_assigner.py` — no-op single-nation side, mixed-nation side, cross-side country collision, faction-country reservation, belligerent ids, and registered-instance identity.
- `tests/squadrons/test_pilotnames.py` — mapped country → its own locale, unmapped/None → fallback, locale caching, **every** mapped locale builds a usable gender-aware Faker (guards the table against typos), and a squadron recruits named pilots from its country locale.

Black + mypy (`game tests`) + full pytest suite all green locally.


geofffranks:
Claude had the following findings on this. Can you please take a look at items 1 + 2, and confirm whether 3-5 are actual issues? Not sure if you'd want to address 6-10 or not.

## Correctness bugs (CONFIRMED)

1. Mirror faction → same Country in both coalitions — game/missiongenerator/countryassigner.py:49
No guard for blue.faction.country.id == red.faction.country.id. _instance() returns same Country object for both primary_blue and primary_red. setup_mission_coalitions then calls add_country(same_instance) on both sides → pydcs raises or DCS makes unloadable .miz. Mirror scenarios (same country id both sides) break.
Fix: guard equal-id case; give red its own instance or route to distinct coalition.

2. Pretense bypasses the whole fix — game/pretense/pretensemissiongenerator.py:155
PretenseMissionGenerator.setup_mission_coalitions overrides base method, never touches country_assigner. Still manually computes belligerents = {p_country.id, e_country.id} (dupes belligerent_ids) and registers only single faction country per side (lines 166-167) instead of country_assigner.blue_countries / red_countries. Every Pretense campaign silently skips per-squadron registration → all aircraft spawn under single faction country. Voiceover fix defeated for all Pretense users, any coalition type.
Fix: call country_assigner.blue_countries/red_countries/belligerent_ids in the override, like base class.

## Correctness bugs (PLAUSIBLE — verify)

3. neutral_country drift — game/missiongenerator/missiongenerator.py:192
belligerent_ids now includes squadron country IDs, but game.neutral_country only checks faction primaries. CJTF-blue + Swiss squadron registers Switzerland in blue; neutral_country still returns Switzerland → add_static_group puts helipads in blue not neutral → breaks DCS capture triggers keyed on neutral membership.

4. Unguarded KeyError in __init__ — countryassigner.py:97
country_dict[cid]() for every squadron country id. Missing id (pydcs version drop, uninstalled mod in save) → KeyError aborts mission gen. Old code only looked up 2 faction ids.
Fix: .get() + skip/fallback with log.

5. faker_for_country lru_cache breaks test isolation — game/squadrons/squadron.py:283
Module-level @lru_cache(maxsize=None) on _faker_for_locale caches for process life. Tests patching Faker/locales get stale pre-patch instance.

## Quality / cleanup (CONFIRMED)

6. faker property not cached — squadron.py:283
Accessed once per pilot in _recruit_pilots range(count) loop → hundreds of faker_for_country calls/campaign, all identical. @property → functools.cached_property, O(pilots)→O(1) per squadron.

7. Test never cross-checks country keys vs pydcs — tests/squadrons/test_pilotnames.py:46
test_every_mapped_locale_generates_a_name only exercises Faker locale construction. Misspelled/stale key ('Czeck Republic') → dead table entry, pilots silently get default names, test still green. Add pydcs country_dict cross-check.

8. Redundant double RNG calls in test — test_pilotnames.py:52
name_male()/name_female() called twice per assertion (isinstance + truthiness). isinstance noise — always str. Simplify to assert faker.name_male().

## Design smells (PLAUSIBLE)

9. Parallel registration tables — game/squadrons/pilotnames.py:25
COUNTRY_FAKER_LOCALES keyed by Country.name string; CountryAssigner keys by Country.id int. pydcs rename → pilots silently revert to default names, no signal, while CountryAssigner keeps working. Two tables, divergent failure modes.

10. Silent faction-primary fallback — countryassigner.py:116
for_squadron falls back to faction primary silently, inconsistent with debug logging on every other skip/collision in __init__. Dynamically-added reinforcement squadron gets wrong country, no log.

geofffranks (changes requested):
See above comment

BradySox:
> Claude had the following findings on this. Can you please take a look at items 1 + 2, and confirm whether 3-5 are actual issues? Not sure if you'd want to address 6-10 or not.
> 
> 
> 
> ## Correctness bugs (CONFIRMED)
> 
> 
> 
> 1. Mirror faction → same Country in both coalitions — game/missiongenerator/countryassigner.py:49
> 
> No guard for blue.faction.country.id == red.faction.country.id. _instance() returns same Country object for both primary_blue and primary_red. setup_mission_coalitions then calls add_country(same_instance) on both sides → pydcs raises or DCS makes unloadable .miz. Mirror scenarios (same country id both sides) break.
> 
> Fix: guard equal-id case; give red its own instance or route to distinct coalition.
> 
> 
> 
> 2. Pretense bypasses the whole fix — game/pretense/pretensemissiongenerator.py:155
> 
> PretenseMissionGenerator.setup_mission_coalitions overrides base method, never touches country_assigner. Still manually computes belligerents = {p_country.id, e_country.id} (dupes belligerent_ids) and registers only single faction country per side (lines 166-167) instead of country_assigner.blue_countries / red_countries. Every Pretense campaign silently skips per-squadron registration → all aircraft spawn under single faction country. Voiceover fix defeated for all Pretense users, any coalition type.
> 
> Fix: call country_assigner.blue_countries/red_countries/belligerent_ids in the override, like base class.
> 
> 
> 
> ## Correctness bugs (PLAUSIBLE — verify)
> 
> 
> 
> 3. neutral_country drift — game/missiongenerator/missiongenerator.py:192
> 
> belligerent_ids now includes squadron country IDs, but game.neutral_country only checks faction primaries. CJTF-blue + Swiss squadron registers Switzerland in blue; neutral_country still returns Switzerland → add_static_group puts helipads in blue not neutral → breaks DCS capture triggers keyed on neutral membership.
> 
> 
> 
> 4. Unguarded KeyError in __init__ — countryassigner.py:97
> 
> country_dict[cid]() for every squadron country id. Missing id (pydcs version drop, uninstalled mod in save) → KeyError aborts mission gen. Old code only looked up 2 faction ids.
> 
> Fix: .get() + skip/fallback with log.
> 
> 
> 
> 5. faker_for_country lru_cache breaks test isolation — game/squadrons/squadron.py:283
> 
> Module-level @lru_cache(maxsize=None) on _faker_for_locale caches for process life. Tests patching Faker/locales get stale pre-patch instance.
> 
> 
> 
> ## Quality / cleanup (CONFIRMED)
> 
> 
> 
> 6. faker property not cached — squadron.py:283
> 
> Accessed once per pilot in _recruit_pilots range(count) loop → hundreds of faker_for_country calls/campaign, all identical. @property → functools.cached_property, O(pilots)→O(1) per squadron.
> 
> 
> 
> 7. Test never cross-checks country keys vs pydcs — tests/squadrons/test_pilotnames.py:46
> 
> test_every_mapped_locale_generates_a_name only exercises Faker locale construction. Misspelled/stale key ('Czeck Republic') → dead table entry, pilots silently get default names, test still green. Add pydcs country_dict cross-check.
> 
> 
> 
> 8. Redundant double RNG calls in test — test_pilotnames.py:52
> 
> name_male()/name_female() called twice per assertion (isinstance + truthiness). isinstance noise — always str. Simplify to assert faker.name_male().
> 
> 
> 
> ## Design smells (PLAUSIBLE)
> 
> 
> 
> 9. Parallel registration tables — game/squadrons/pilotnames.py:25
> 
> COUNTRY_FAKER_LOCALES keyed by Country.name string; CountryAssigner keys by Country.id int. pydcs rename → pilots silently revert to default names, no signal, while CountryAssigner keeps working. Two tables, divergent failure modes.
> 
> 
> 
> 10. Silent faction-primary fallback — countryassigner.py:116
> 
> for_squadron falls back to faction primary silently, inconsistent with debug logging on every other skip/collision in __init__. Dynamically-added reinforcement squadron gets wrong country, no log.
> 
> 
> 
> 
> 
> 

Makes total sense - I completely cut pretense out of my fork. Trusty Claude I think addressed your findings?

geofffranks:
It looks like you merged https://github.com/bradyccox/414Ret/pull/529 to your main branch rather than the feat/per-squadron-country branch, can you push that up to the branch this PR is based on?

geofffranks:
Ok, took another look and claude said it's mostly fixed, but found a couple new things:

1. Pretense package flights still spawn under faction country — pretenseaircraftgenerator.py:918 (CONFIRMED)
generate_packages passes faction-level p_country to create_and_configure_flight, not country_assigner.for_squadron(flight.squadron). setup_mission_coalitions registers the national countries correctly, but flights still fly under CJTF. Per-squadron voiceover still defeated for Pretense package flights — the original Pretense bug (#3) only half-fixed. This is the one that matters.

2. Wrong docstring on pydcs equality — game.py:232 (CONFIRMED)
Comment claims "pydcs Country uses identity equality, so a set of instances never detects a match." False — Country.__eq__/__hash__ are by .id. Misleads future container/dedup logic. The fix works, but the stated reason is wrong.

3. Spurious debug log on mirror match — countryassigner.py:68 (CONFIRMED)
When blue_id == red_id, blue faction-country squadrons hit the red-guard continue first → logs "falls back to primary_blue" even though it resolves correctly. Noise obscures real fallbacks in logs. Cosmetic.

4. Test still imports private _faker_for_locale — test_pilotnames.py:12 (CONFIRMED). Prior #6, unaddressed. Rename → ImportError not test-fail.

PLAUSIBLE

- 5. neutral_country init-ordering — evaluated before configure_default_air_wing; stored neutral_faction.country (e.g. Switzerland) can diverge from live property after auto-gen squadrons appear → neutral statics/capture triggers misrouted — game.py:147
- 6. neutral_country now walks all squadrons every call, invoked 8+×/mission-gen → O(n·8) — game.py:238
- 7. Pretense blue_ids/red_ids set comprehensions re-alloc data already in CountryAssigner — pretensemissiongenerator.py:179

geofffranks (approved):
LGTC

geofffranks:
thanks for all the fixes!

BradySox:
> thanks for all the fixes!

Thanks for helping find the bugs- appreciate it

Starfire13:
This would actually be quite handy for my upcoming Syria campaign where I have a mix of Israeli and American squadrons in the same faction. :)

game/game.py
```diff
@@ -226,14 +226,26 @@ def air_wing_for(self, player: Player) -> AirWing:
...
+        ``__hash__`` are by ``id``).
+        """
+        ids_in_use = {self.red.faction.country.id, self.blue.faction.country.id}
+        for coalition in (self.blue, self.red):
+            for squadron in coalition.air_wing.iter_squadrons():
+                ids_in_use.add(squadron.country.id)
+        for candidate in (UnitedNationsPeacekeepers, Switzerland, USAFAggressors):
+            if candidate.id not in ids_in_use:
+                return candidate()
+        return USAFAggressors()
```
Druss99: do we have a test case for if all 3 of the candidates are in use?

game/squadrons/squadron.py
```diff
@@ -275,7 +280,11 @@ def return_from_leave(self, pilot: Pilot) -> None:
 
     @property
     def faker(self) -> Faker:
-        return self.coalition.faker
```
Druss99, line 278: Is coalition.faker used anywhere else? If not we should remove it, if so then we need to evaluate unifying the two.

Druss99 (changes requested):

game/game.py
BradySox: I think Claude grabbed a red flag campaign damn it. Lemme see

BradySox: We do now — and writing it surfaced a real bug: the final fallback returned `USAFAggressors()` even when claimed. With Aggressors as the red faction and a blue CJTF fielding UN and Swiss squadrons, all three preferred neutrals are in use, and the old tail put one country on two coalitions (the unloadable .miz this property exists to prevent). The tail now scans the full pydcs country list for an unclaimed nation, and raises rather than hand back a claimed one if even that fails. New `tests/test_game_neutral_country.py` covers the preferred pick, the squadron-claimed fall-through, and the all-three-claimed scan.

game/squadrons/squadron.py
BradySox: Audited: the only live consumer was `Squadron.faker`'s fallback — `Game.faker_for` had zero callers. Both are removed (`Coalition.on_load` with them; recreating the faker was its whole job), and the fallback now builds from the faction's locale list via a cached `faker_for_locales` in `pilotnames.py`, so there is exactly one construction path beside `faker_for_locale`. The faker was never persisted, so no save impact.

BradySox:
@Druss99 I assume "resolving" is the correct way to do it, they should be addressed for both PR you messaged about today

Druss99 (approved):

Druss99:
Can you update the changelog? Then we can merge

Druss99 (approved):

PR #856, opened 2026-07-06
Update CH USA pack to 1.5.0

juanjux:
Update CH USA pack to 1.5.0, which should fix the Patriot batteries that were not being added if you had the mod updated to the last version. Also, compatibility with older saves.


Druss99:
This is because the mod needs to be updated to the latest version in retribution and the old unit IDs/names would need to be migrated, see the wiki page regarding adding a new mod and this PR for reference: https://github.com/dcs-retribution/dcs-retribution/pull/641

juanjux:
> This is because the mod needs to be updated to the latest version in retribution and the old unit IDs/names would need to be migrated, see the wiki page regarding adding a new mod and this PR for reference: #641

Thank you, I will take a look and update the PR upgrading the CH USA mod.

juanjux:
> This is because the mod needs to be updated to the latest version in retribution and the old unit IDs/names would need to be migrated, see the wiki page regarding adding a new mod and this PR for reference: #641

Done! PTAL again.

Druss99:
Since there are only the patriot sites renamed, did you do a pydcs export of the mod? I would've expected a bit more in terms of changes

juanjux:
> Since there are only the patriot sites renamed, did you do a pydcs export of the mod? I would've expected a bit more in terms of changes

I didn't (learning as I go) but now I did, please take a look when you can to see if something else is amiss, included some changes for backward compatibility with previous saves.

game/dcs/groundunittype.py
```diff
@@ -119,6 +119,15 @@ def _migrator() -> Dict[str, str]:
             "[CH] Tor M2 SHORAD": 'SAM SA-15 Tor M2 "Gauntlet" [CH]',
             "[CH] Tor M2M SHORAD": 'SAM SA-15 Tor M2 "Gauntlet" [CH]',
             "[CH] Iskander-M SRBM": "SRBM 9K720 Iskander HE [CH]",
+            # CH USA pack 1.5.0 export refresh: units renamed/removed in the mod ->
+            # migrate old saves to the closest surviving unit (by display name).
+            "[CH] M142 HIMARS (GLSDB)": "[CH] M270A1 MLRS (GLSDB)",
```
Druss99: A bunch of currenthill mod units have been added to DCS as he partnered with ED. Most of the units that were removed from his Russia pack were removed because some variant of them was added to DCS. The DCS units all have CHAP in the name. Here's an example: https://github.com/dcs-retribution/dcs-retribution/blob/dev/resources/units/ground_units/CHAP_M142_ATACMS_M48.yaml

Let's migrate these to units in the base game (the CHAP units) where possible

Druss99 (changes requested):

game/dcs/groundunittype.py
juanjux: Interesting, didn't knew that, great that they are taking from the community (I sent them a bugfix for a lua file and the bug is still in the last release). I have updated the PR, basically all the HIMARs and some other hardware are now in the game with the CHAP prefix as you correctly guessed, PTAL.

Druss99: I see you removed the redundant units from the faction file but we still need to migrate the old mod units to the new native DCS units here

game/persistency.py
```diff
@@ -219,6 +220,31 @@ def _handle_ch_russian_assets(self, module: str, name: str) -> Any:
...
+        FMTV/M-ATV trucks, B-21) -> map old saves to the closest current class."""
+        if module != "pydcs_extensions.usamilitaryassetspack.usamilitaryassetspack":
+            return None
+        from pydcs_extensions.usamilitaryassetspack import usamilitaryassetspack
+
+        if name.startswith("MIM104_"):
+            return getattr(usamilitaryassetspack, "CH_" + name, None)
+        renames = {
+            "B_21": "CH_B_21",
+            "M142_HIMARS_GLSDB": "CH_M270A1_GLSDB",
```
Druss99: If I'm reading the changes correctly, everything this line and below needs to move into the groundunit file to be migrated to the native DCS unit.

IIRC if we migrate within the mod we do it here and migrating from the mod to native DCS needs to happen in the other file. Most of these are already handled in your latest update via the groundunit file but some are not

game/dcs/groundunittype.py
juanjux: Display names migrated to the ones shown by DCS in the last commit. Note that the CHAP is in the id, not in the display name (if I'm not wrong, which I could be since it's the first time I do this).

Druss99 (changes requested):

game/dcs/groundunittype.py
juanjux: PS: I am adding comment with the id string so the mapping is clearer.

Druss99:
Also one last note, due to all the migrations I recommend creating a campaign using this mod and one of the factions off the dev branch and then trying to load the save and click take off using this PR. That should help find any migration typos or misses

juanjux:
> Also one last note, due to all the migrations I recommend creating a campaign using this mod and one of the factions off the dev branch and then trying to load the save and click take off using this PR. That should help find any migration typos or misses

Yeah I did that, but that was before the latest changes from your review, need to so it again (soon, little time these latest days).

game/persistency.py
juanjux: 🤦 Check the [latest commit please](https://github.com/dcs-retribution/dcs-retribution/pull/856/changes/defc9abae8163a676080085b59ab53e4cffc57e7), I hope nothing else is amiss!

Druss99 (approved):

Druss99:
Sorry I missed it in the other reviews, could you please update the UI to note the new version here: qt_ui/windows/newgame/WizardPages/QGeneratorSettings.py

PR #857, opened 2026-07-08
fix(carcass): dedup respawned-static carcasses by type+position

geofffranks:
Game.add_destroyed_units appended every destroyed-static position from the debrief with no dedup. Statics that Retribution respawns alive every mission at a fixed spot — FARP fuel/ammo depots (FARP_Fuel_Depot, FARP_Ammo_Dump_Coating), the motorpool Garage_A depot — therefore stacked a new hidden dead carcass at that spot each time they were bombed, and a live respawned unit z-fought the old wreck. Growth is unbounded over a campaign; generate_destroyed_units re-spawns every entry each mission (gated by perf_destroyed_units). Perf/cosmetic — no economy or front-line effect.

##    Steps to reproduce

    1. Enable perf_destroyed_units. Play a campaign with a FARP (fuel/ammo depots present).
    2. Bomb the FARP fuel/ammo depot; end the mission.
    3. Repeat over several turns (the depot respawns alive each mission).
    4. Inspect the save: game.get_destroyed_units() holds N entries at the same (type, x, z) — one per mission bombed — and each generated .miz spawns N coincident hidden dead statics, one of them clipping the live respawned depot.

## Fix

    1. Dedup carcasses by position — game/game.py
    2. Suppress carcasses under respawned live units

      Layering stays clean: LiveUnitIndex is a leaf; game.py and missiongenerator.py both depend on it, with no new game → missiongenerator import.

      Behavior for genuinely one-time destroyed statics is unchanged (they were already exactly one wreck, and no live unit stands on them). 22 new tests; full suite green.

### DIFF
--- a/game/game.py
+++ b/game/game.py
@@ -33,6 +33,7 @@ from .infos.information import Information
 from .lasercodes.lasercoderegistry import LaserCodeRegistry
 from .profiling import logged_duration
 from .settings import Settings
+from .spatialindex import LiveUnitIndex
 from .theater import ConflictTheater, Player
 from .theater.bullseye import Bullseye
 from .theater.theatergroundobject import (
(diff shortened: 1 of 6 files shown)


juanjux (approved):

juanjux:
LGTM (but not a fan of the size of Claude's comments on the code, I always tell mine to shorten them), love the huge amount of tests.

PR #859, opened 2026-07-08
Strikeable motorpool depots

### LINKED ISSUE #655: Motorpool Support

Right now the only way to slow/stop the flow of units is by destroying a factory which neuters the opfor.

Instead of reserve units just not spawning, it would be nice if control points could have a motorpool object where all the units not on the frontline would spawn. That way you could target those units with a strike to destroy the opfor reserve and make them spend more income to replace their reserve units. This would be more fun than destroying a factory and completely eliminating the opfors way of resupplying units and might allow money to play more of a factor in campaigns

### Additional context

_No response_

geofffranks:
Adds a strikeable **motorpool** ground object at control points where reserve ground units spawn, so a player can attrit an enemy's armor reserve directly by bombing the depot instead of only meeting it at the front line.

### What it does
- **New `MotorpoolGroundObject`** — a per-turn live projection of the reserve slice of a control point's `Base.armor`, rendered as a strikeable vehicle group. Gated on authored `Fortification.Garage_A` preset locations; symbol is a maintenance facility (distinct from armor groups).
- **1:1 grind, no economy** — each destroyed motorpool vehicle decrements `base.armor` by one, forcing repurchase next turn. Pure removal: no income, and depot strikes do **not** shift the front line (motorpool losses are tracked separately from front-line casualties).
- **Passive reserves** — rendered vehicles are parked (ROE weapon-hold, no waypoints, `player_can_drive=False`): they don't move or return fire, but still register kills.
- **Depot presentation** — the map shows a present depot icon (never "destroyed", even before mission generation or after all vehicles are killed); in-mission spawns a Garage A building plus the parked vehicles.
- **Codename** — motorpools get a unique codename like other TGOs.
- **Debrief display** — motorpool losses now appear in the post-mission debrief (faction overview "Motorpool units lost" row + detailed "`<type>` from motorpool" rows).
- **Save migration** — existing campaigns with authored motorpools get the TGO injected on load.

### Settings
`motorpool_enabled` (on) and `motorpool_spawn_cap` (10), both user-visible.

### Known limitations
- Pretense mode unsupported (no loss reconciliation there); normal campaigns only.
- Multiple motorpools on one CP share the single reserve pool (round-robin).
- Depot is a respawned destructible static, so a hidden dead-carcass accumulates at its position over many missions — systemic to all Retribution-respawned statics, no economy/front impact, tracked separately via #857.

### Testing
CI gates green (black, mypy game+tests, pytest). In-game acceptance validated: map icon, in-mission building + parked vehicles, grind/repurchase, no front-line shift, spawn-cap, save load, codename, debrief display.

Closes #655


Starfire13:
What are the dimensions of the motor pool parking lot? Are they parked in a grid? Can campaign designers manually designate where this parking lot is placed?

geofffranks:
Grid of vehicles at 12 m spacing, 5 columns wide, growing in +x/+y from the TGO position (motorpoolpopulator.py:16-17, _make_unit :106-119). Default motorpool_spawn_cap = 10 → 5×2 → footprint ≈ 48 m × 12 m (4 gaps × 12 m wide, 1 gap tall). Cap is user-configurable, so at higher caps it just adds rows (each +5 vehicles = +12 m depth).

The depot building (Garage A) sits offset -50 m, -50 m from the grid origin (motorpoolgenerator.py:20,57-60) — opposite direction from grid growth, so it never collides with a parked unit (DCS silently drops overlapping spawns).

Placement is fully author-controlled — not auto-generated. Designer drops a Fortification.Garage_A static in the campaign .miz. Loader (mizcampaignloader.py:289-294) matches on exact type id, assigns each to the nearest control point's preset_locations.motorpools (:693-695). That static's position is the grid origin. No authored Garage A → no motorpool at that CP (start_generator.py:536-544). So designers control both whether and where per CP.

Starfire13:
Is the orientation of the garage building used to determine orientation of the parking grid?

geofffranks:
No, not currently. The grid is always world-axis-aligned: it grows north (+x) for each column and east (+y) for each new row of 5. The Garage A heading rotates the individual vehicles' facing and the depot building model, but not the grid axes. At a non-cardinal heading you'll get angled vehicles in a still-N/E grid.  But we can make that happen if you want. 

---

Campaign designer's guide (for posterity + the wiki)

1. **The trigger:** Drop a Fortification.Garage_A static into the campaign .miz. That exact static type (and only that type) is what the loader scans for. No Garage A → no motorpool.

2. **Where vehicles spawn:** The static's position is grid slot 0 — the first vehicle spawns exactly on it. The grid then extends north and east from there (5 vehicles per row at 12 m, rows step 12 m east; default cap 10 = a 48 m N × 12 m E block). Place the static at the SW corner of where you want the lot.

3. **Leave room SW:** The depot building spawns 50 m south + 50 m west of the static (opposite the grid, so it never overlaps a vehicle). Keep that corner clear of other statics/terrain.

4. **Heading:** Sets vehicle facing + building facing. Does not rotate the grid (see above).

5. **Which CP owns it:** Each Garage A is auto-assigned to the nearest control point (objective_info → closest). If two CPs are close, place the static clearly nearer the intended one, or it can bind to the wrong CP.

6. **Multiple per CP Allowed:**. Several Garage A statics on one CP share that CP's single armor reserve pool, dealt round-robin. Two motorpools ≠ double the reserve — it's the same pool split across locations.

7. **Settings that gate it:** motorpool_enabled (default on) and motorpool_spawn_cap (default 10). Cap 0 or disabled → nothing spawns. Cap only limits rendered/strikeable vehicles, chosen proportionally across reserve types.

8. **Existing campaigns:** Save migration injects the TGO on load, so adding a Garage A to an in-progress campaign's .miz works without a new game.

Starfire13:
I think it would be easier to get a good fit if the building's orientation mattered. That's how it's set up for the resource site placement (factories and such). In fact, IIRV one of the long-term issues we currently have is that FOB/FARP buildings were not set up to take orientation into account.

geofffranks:
ok, orientation of the building now rotates the orientation of the motorpool grid too now.

Starfire13:
Sweet. By the way, when this gets merged, we will probably want to place a visual representation of how to set it up on the wiki. Just a description with screenshot showing the grid of vehicles in relation to the building should be enough.

Starfire13:
One other thing I just thought of. Are these ground units set up as statics like the untasked aircraft at airfields, or regular ground units with ROE set to none? I wonder if there's a difference in terms of base capture.

geofffranks:
> One other thing I just thought of. Are these ground units set up as statics like the untasked aircraft at airfields, or regular ground units with ROE set to none? I wonder if there's a difference in terms of base capture.

yeah this is almost definitely an issue. i'm going to test out check that'll pop up loudly when a campaign has a motorpool within 3km of the base.

Starfire13:
Gotcha. For the moment, I am avoiding the issue on the campaign I am designing by placing the motor pool outside the capture radius of the control points.

game/missiongenerator/motorpoolgenerator.py
```diff
@@ -0,0 +1,93 @@
...
+        # Motorpool units register as motorpool losses (see generate), NOT as
+        # theater objects — theater-object deaths do not touch base.armor.
+        return
+
+    def enable_eplrs(self, group: VehicleGroup, unit_type: Type[VehicleType]) -> None:
+        # Unmanned parked vehicles must not broadcast datalink/targeting data to
+        # coalition AI, so suppress the EPLRS task the base generator would add.
+        return
+
+    def _passivate(self, group: VehicleGroup) -> None:
```
Druss99: maybe _set_passive instead of this made up word haha

game/settings/settings.py
```diff
@@ -532,6 +532,30 @@ class Settings:
...
+            "Destroying reserves forces the owner to repurchase."
+        ),
+    )
+    motorpool_spawn_cap: int = bounded_int_option(
+        "Maximum motorpool vehicles per turn",
+        page=CAMPAIGN_MANAGEMENT_PAGE,
+        section=GENERAL_SECTION,
+        default=10,
+        min=0,
+        max=50,
```
Druss99: I see the cap is 50 here. Regarding the size of the motorpool can it fit 50 units? Is the max size of the grid capped? From what I understand it seems like the size is capped at 5 rows of 10 units each right?

Druss99 (changes requested):
I think we should probably allow the autoplanner to plan strike or cas packages against the motorpool and I don't think I see that logic here

Starfire13:
Not CAS I think, since we would want these to be statics and not actual ground units for performance reasons, right? Like the parked aircraft at airfields. This would basically be designed like OCA/Aircraft for ground units?

Edit: thinking about it a bit more, the mission type is similar to OCA/Aircraft in terms of aim (fly to enemy airfield, bomb enemy units) but it's also different because while OCA/Aircraft involves a "search and destroy" type attack, blowing up some trucks in a parking lot in a fixed location can probably be much more simply handled as a strike-type tasking.

Druss99:
Do they have a static version of every normal vehicle unit?

Starfire13:
That, I don't know. But if they don't, and it turns out the performance savings are significant for using statics, perhaps we should either restrict the motor pool units to available statics, or at least make that an option.

geofffranks:
> I think we should probably allow the autoplanner to plan strike or cas packages against the motorpool and I don't think I see that logic here

Yeah i'll update it to allow strike + bai from autoplanner. Good catch!

> maybe _set_passive instead of this made up word haha

Y u no like slop? Will fix.

>  From what I understand it seems like the size is capped at 5 rows of 10 units each right?

Yeah, right now this means the grid would max at 48m wide by 108m deep. maybe we drop this to 5x5 + 25 units capped to get a 48m square?

Regarding the discussion on statics - all the OCA aircraft being spawned are not statics, neither are the motorpool units. The aircraft is all cold-start + set with no tasking, and then set as uncontrolled.  The motorpool units work the same. uncontrolled set to true, weapons hold, drivable=false. What are the performance concerns here?

Starfire13:
Ah I had thought the aircraft were supposed to be statics but maybe that got changed at some point. The concern about performance is because all units, even if stationary and with no ROE, suck up CPU cycles. So if you have a large number of parked ground units, it can significantly bump up your unit counts on campaigns with many control points and hurt performance.

Druss99:
Yeah let's cap it at 25 units max and let's continue to spawn real units, the option is there to disable it for performance for those who need to

geofffranks:
ok, just pushed up:

- cap at 25 units
- warning that pops up when you've configured a motorpool within 3km of the control point
- autoplanner will plan on motorpools

I believe that's the remaining open items, but let me know if i missed any. 🤞

geofffranks:
(minus updating the wiki post-merge)

Starfire13:
Why 3 km when the standard DCS capture radius is 5 nautical miles?

geofffranks:
Yeah i was a little surprised AI thought it was so short, but it was (and remains) adamant that this was the radius retribution uses. I was also under the same impression you were, but after about an hour and a half of pushing back and investigating:

Retribution injects a 3km trigger zone:
radius: https://github.com/dcs-retribution/dcs-retribution/blob/ef576accb08ff6dc85a42d31a89de84223607c9b/game/theater/controlpoint.py#L102
adding the trigger: https://github.com/dcs-retribution/dcs-retribution/commit/e273e9301279771cc1a24c9cdb6c7840a4ec86e6

Trigger zone logic:
https://github.com/dcs-retribution/dcs-retribution/blame/ef576accb08ff6dc85a42d31a89de84223607c9b/game/missiongenerator/triggergenerator.py#L183-L217

Additionally, it sounds like 5nm isn't necessarily accurate for the base capture in dcs? 
https://forum.dcs.world/topic/96947-event-handling/ (really old though). But DCS base captures trigger  S_EVENT_BASE_CAPTURED, and the only reference to that in retribution code is our copy of moose, in code that i'm pretty sure we don't use.  

We went on a wild ride through time - https://github.com/dcs-liberation/dcs_liberation/issues/2566 (hello 2022's Starfire!), and it looks like liberation changed the behavior shortly afterwards (probably because of all the problems seen with infantry vs armor units and spawned inside the zone vs outside, and general ED bugs being ever-present problems): 

https://github.com/dcs-retribution/dcs-retribution/commit/8292da414
https://github.com/dcs-retribution/dcs-retribution/commit/4b0215c16

Starfire13:
Ah yes. I had forgotten we needed our own capture trigger for FOBs/FARPs. I had assumed it would just use the same 5NM radius that DCS does though. I guess not! Good work on the archeology.

game/settings/settings.py
```diff
@@ -532,6 +532,30 @@ class Settings:
...
+            "Render each control point's not-yet-deployed reserve armor as a "
+            "strikeable motorpool (only where the campaign authored one). "
+            "Destroying reserves forces the owner to repurchase."
+        ),
+    )
+    motorpool_spawn_cap: int = bounded_int_option(
+        "Maximum motorpool vehicles per turn",
+        page=CAMPAIGN_MANAGEMENT_PAGE,
+        section=GENERAL_SECTION,
+        default=25,
```
Druss99: Can we default to 10?

Druss99 (changes requested):
Also let's update the change log

game/settings/settings.py
geofffranks: Will do. Also there's a bug that might be annoying in campaign dev. Right now it's warning pops up whenever the campaign loads with a motorpool within the capture zone, but after that, the saved campaign never updates the motorpool location if you change the campaign template .miz file, and a new campaign is needed to propagate the change in. @Starfire13 how annoying is that going to be in the campagin dev workflow? I find it pretty annoying for me since I added it to the campaign i'm playing through with a friend and moved it close to validate the popup, but idk if fixing that long term is worth it if typically you build the .miz and rebuild campaigns until the starting point is what you want?

Druss99: no that is intended, you have to start a new campaign to get the template changes. We definitely don't want to update a campaign from the template on load since you could brick a campaign if someone overhauls it

Druss99 (approved):

game/settings/settings.py
Starfire13: Yeah, what Eclipse said. Once a campaign is generated, it is divorced from campaign template alterations. I don't think it's too much of an issue anyway. Main thing is to document in the wiki how the structure for this feature ought to be placed and how it ought not to be.

geofffranks:
Wiki page is created -https://github.com/dcs-retribution/dcs-retribution/wiki/Motorpools

PR #868, opened 2026-07-13
Naval: fleets actively defend (RED alarm + weapon-free ROE), not passive

juanjux:
Enemy (and friendly) ship groups were generated with only the default alarm state, so when the performance **red alert state** setting is off they sit passive — radars off, not firing — even when the enemy flies right over them.

This forces **RED alarm** on ships (they should never start dark, unlike ground SAMs which do so on purpose for Skynet IADS) and sets **weapon-free ROE**, so a fleet fires autonomously on any target that enters weapon range: SAMs on aircraft, anti-ship missiles / guns / torpedoes on enemy ships (submarines fire torpedoes). Ships still hold position (no attack waypoints) — this is aggressive defense, not a hunt.

Engagement is **option-driven** (ROE + alarm), not task-driven. Note the comment on `set_ship_engagement`: an `EngageTargets` task is **air-only and invalid for a ship controller** (DCS `me_action_db` offers ships only `NoTask`), and feeding it to the naval AI crashes DCS (`ACCESS_VIOLATION` in `AI::ControllerStack::start`) — so it is deliberately not used.

One file (`tgogenerator.py`); no behaviour change for ground groups.

### DIFF
--- a/game/missiongenerator/tgogenerator.py
+++ b/game/missiongenerator/tgogenerator.py
@@ -40,6 +40,7 @@ from dcs.task import (
     EPLRS,
     FireAtPoint,
     OptAlarmState,
+    OptROE,
 )
 from dcs.terrain import Airport
 from dcs.translation import String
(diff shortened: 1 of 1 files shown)


Druss99:
Just making sure this won't make the destroyers fire tomahawks at ground targets right? I think that's a no but I want to be sure

juanjux:
> Just making sure this won't make the destroyers fire tomahawks at ground targets right? I think that's a no but I want to be sure

@Druss99 I am currently playing the Marianas USA vs China campaign, and they only fire RGM-84 to other ships (and the Chinese ones, both default and CH China, also fire only anti ship missiles).

Druss99 (approved):

PR #870, opened 2026-07-15
Airboss: add S3/A6 recovery tanker selection dropdown

red-one1:
Add a plugin-option dropdown to choose between the S-3B and Heatblur A-6E as the Airboss recovery tanker (default S3). Introduces a string/choice plugin-option type (possibleValues) rendered as a combo box, with values injected as quoted Lua strings. Adds the A-6E tanker spawn (buddy-refuel loadout) and wires the CVN dispatch to the selected type.
image

### DIFF
--- a/game/plugins/luaplugin.py
+++ b/game/plugins/luaplugin.py
@@ -58,10 +58,17 @@ class PluginSettings:
 
 class LuaPluginOption(PluginSettings):
     def __init__(
-        self, identifier: str, name: str, min: Any, max: Any, value: Any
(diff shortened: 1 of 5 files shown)


Druss99:
Can you take your pydcs PR out of draft if it's ready so we can merge it? After that you'll need to update the pydcs ref in this PR so the A6 is usable as a tanker.

Also can you add dashes to the S-3 and A-6

red-one1:
Pydcs#33

PR #872, opened 2026-07-16
Ship-launched cruise missile strikes: Tomahawk/Kalibr shore attack with a finite campaign magazine

BradySox:
Warships that carry land-attack cruise missiles (the vanilla Arleigh Burke's Tomahawks and Ticonderoga's TLAMs, the CurrentHill packs' Kalibr LACM/CMP hulls) can now fire real cruise missile strikes at shore targets, driven by a new `cruisemissiles` plugin. Both settings default **off**, so existing campaigns are untouched.

## What it does

- **F10 call-for-fire**: each coalition gets a "Cruise Missile Strike" menu — put a map marker on a shore target and the nearest capable ship with stock ripples a salvo at it via a scripted `FireAtPoint` push carrying the cruise-missile weapon flag (2097152), i.e. the same mechanism the mission editor uses. Putting just a number in the marker's text (`6` or `#6`) sizes the salvo; anything else fires the default.
- **Auto raids** (separate setting): each side commits at most one raid per turn at its highest-value reachable enemy ground object — command/comms first, then war industry — announced to the defender only as a LAUNCH WARNING.
- **Finite campaign magazine**: DCS silently rearms ships every mission, so the campaign persists a per-ship-group missile stock (`game.cruise_missile_magazines`) with **no rearm**. A "Magazine status" F10 menu shows what's left.
- **Real weapons, no scripted kills**: the missiles come from real, tracked, sinkable ships — kills record natively at debrief, point-defense SAMs can intercept them, and a sunk shooter fires nothing. The plugin owns no kills and no spawns. Fully symmetric between coalitions.
- **Surfaced in the UI, side-gated**: the mission briefing gets a CRUISE MISSILES section (the player's tasked raid + each friendly launcher's remaining magazine + the call-for-fire how-to), the ground-object dialog shows a friendly naval group's remaining missiles, and the debriefing window lists missiles expended per group (post-debit remainder for friendly groups only). Enemy residual stock is never shown anywhere.

## How the magazine stays honest

The plugin mirrors expenditure into a new `cruise_missiles_state` channel in the state file (`{group=, fired=}` per launching ship group). Python debits the persisted magazine **only** from that report, at the turn boundary (`commit_cruise_missiles` in the mission results processor) — generation never debits, so re-generating a mission cannot double-count, and pre-feature state files simply parse as empty.

## Files

- `game/cruisemissiles.py` — curated LACM hull set (DCS/pydcs expose no per-ship weapon taxonomy), magazine seeding/debit, auto-raid planner
- `game/missiongenerator/cruisemissileluadata.py` + `luagenerator.py` wiring — emits `dcsRetribution.cruiseMissiles` only when the setting is on and a live launching group exists
- `resources/plugins/cruisemissiles/` — the plugin (auto raids, F10 menus, expenditure mirroring)
- `game/debriefing.py`, `game/sim/missionresultsprocessor.py`, `resources/plugins/base/dcs_retribution.lua` — the `cruise_missiles_state` debrief channel
- `game/settings/settings.py` — `cruise_missile_strikes` + `cruise_missile_auto_raids` (Mission Generator page, both default off)
- `game/missiongenerator/briefinggenerator.py` + `resources/briefing/templates/briefingtemplate_EN.j2`, `qt_ui/windows/groundobject/QGroundObjectMenu.py`, `qt_ui/windows/QDebriefingWindow.py` — the UI surfacing (all driven by tested helpers in `game/cruisemissiles.py`)
- `tests/test_cruisemissiles.py`, `tests/missiongenerator/test_cruisemissileluadata.py`, changelog entry

## Validation

On this branch (based on current `dev`, ef576acc): `pytest tests` 249 passed / 2 skipped / 0 failed; `black --check .` clean; `mypy game` and `mypy tests` clean; both Lua files compile on Lua 5.1; the briefing template smoke-renders with and without cruise-missile data (section hides when empty) and both modified Qt modules import clean.


Starfire13:
Missile fire from Scuds and such is randomised within a window post mission start so that not all the missiles are firing at the same time and killing the framerate. We could adopt that same approach for this, so it's not firing exactly 240 seconds after mission start. Would also benefit campaigns that have more than one naval group (or that have naval groups on both sides) set up to fire missiles, to cut down on the framerate loss.

BradySox:
> Missile fire from Scuds and such is randomised within a window post mission start so that not all the missiles are firing at the same time and killing the framerate. We could adopt that same approach for this, so it's not firing exactly 240 seconds after mission start. Would also benefit campaigns that have more than one naval group (or that have naval groups on both sides) set up to fire missiles, to cut down on the framerate loss.

Great point I only test fired 8 total, and when spread across 4 Burkes you don't notice, but with with the option of you being able to just dump the full 96 at one target, I totally see the reason. it should be done now 

"Each planned raid now schedules independently at a random moment inside
[raidDelayMinS, raidDelayMaxS] (defaults 240-900 s; the legacy
raidDelayS option is still honored as the window open). Campaigns with
several launching naval groups -- or groups on both sides -- spread
their salvos across the window instead of rippling together, and the
defender's LAUNCH WARNINGs arrive per raid rather than as one blast."

BradySox:
One last piece before leaving draft: the **defender launch wake**. Flight testing found nothing in DCS ever wakes a defender for a cruise missile on its own — ALARM AUTO only trips on aircraft, and IADS scripts scan units, never weapon objects — so raids flew in unopposed past alive point-defense SAMs 250 m from the aimpoint, while naval SAMs (ships are always hot) intercepted the same salvos natively. Every launch now sets opposing ground AD groups within a radius of the aimpoint to alarm RED for the estimated missile flight window, then stands them back down to AUTO (alarm state only — never emission toggling; a Skynet-managed site keeps its own emission loop and may re-dark, which is the IADS engine's call). Options `defenderWake` (on) / `defenderWakeRadiusNm` (8) / `defenderWakeExtraS` (300). Lua compiles clean under 5.1; the full loop (auto raid, F10 call-for-fire with marker-text salvo sizing, magazine debit across turns) has been flight-tested end-to-end on our side. Marking ready for review.

Ramius007:
Quick reminder, that unlike Kalibrs, Tomahawks use same launchers as air defence missiles.

BradySox:
> Quick reminder, that unlike Kalibrs, Tomahawks use same launchers as air defence missiles.

Good catch, and worth being explicit that the model doesn't account for it. The magazine here (cruise_missile_magazines, 24 for a Burke) is a standalone campaign-level counter for the land-attack raid feature — it has no relationship to the ship's actual DCS weapon loadout or its SM-2/SM-6 stock, which stay exactly what they've always been, untouched by this PR. So today there's no in-sim tradeoff: a Burke can fire its full 24-missile magazine across a campaign and its self-defense missiles are never affected, which isn't quite right against a real Mk41-cell ship.

Modeling actual VLS cell-sharing (debit self-defense capacity when you spend it on Tomahawks) is more scope than this PR — it'd need to reach into whatever tracks the ship's own SAM stock, which today is just DCS's native loadout, not something the campaign layer touches at all.

PR #873, opened 2026-07-16
Culling: keep scenery-objective kill tracking in culled regions

BradySox:
Found by code inspection while investigating a field report of kills silently vanishing at debrief in culled regions.

## The bug

A culled TGO skips *everything* in `GroundObjectGenerator.generate`, including `add_trigger_zone_for_scenery` — but the map buildings backing a scenery objective exist whether or not the campaign spawns anything. So a scenery objective in a culled region is a bombable, visibly-collapsing target whose death is **never recorded**: no `MapObjectIsDead` rule exists, the kill never reaches `dead_events`, and the strike vanishes at debrief with no trace. Skynet also loses its command stand-in for culled IADS scenery nodes.

## The fix

The scenery apparatus — trigger zone, kill-tracking trigger rule, one static infantry — costs effectively nothing, so it now generates regardless of culling. A culled TGO's real spawnable statics, vehicles, and ships are still skipped: culling keeps its performance win.

## Tests

`tests/test_culling.py` locks both halves of the contract: culled scenery keeps its kill tracking; culled spawnable statics still never spawn.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -22,6 +22,7 @@
 
 
 ## Fixes
+* **[Mission Generation]** Scenery objectives keep their kill-tracking trigger zones (and Skynet command stand-ins) even when performance culling skips the rest of the objective. The map buildings backing a scenery objective exist whether or not the campaign spawns anything, so bombing one in a culled region used to visibly destroy a target whose death was never recorded at debrief.
 * **[Plugins]** Fix the escort leash never running (DCS has no `Group.getByID`; look the group up by name via mist), so escorts are actually held to their engagement range.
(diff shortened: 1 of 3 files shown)


Starfire13:
My understanding is that any map object strike target in a culled region that is the mission objective of a package is not actually culled. Is that not the case?

BradySox:
Oh my original comment did not post- @Starfire13 I think this is the mystery of the Strike targets not tracking across some maps. The "M4 man" that will sometimes be on strike targets and sometimes not. Could be completely off base, but its a starting place for scenery objects not tracking

PR #874, opened 2026-07-16
Curated carrier comms: stable, hull-matched CV Operations Data

BradySox:
## What

DCS auto-generates the **"CV Operations Data"** kneeboard page (and the briefing's carrier data) straight from the mission file. The generator currently feeds it allocator output, so every campaign's boat reads like this:

```
Callsign: 0796 | CVN-71 Theodore Roosevelt   <- the theater-unit id prefix on the page
ATC:      346.000 MHz                        <- random, re-rolled EVERY turn
TACAN:    1 X  TRU                           <- first-free channel + random.choice ident, re-rolled every mission
ILS:      1
Link 4:   255.0 MHz                          <- a random inter-flight UHF
```

The page can't be restyled (DCS renders it), so this PR feeds it a real boat card instead, the way the pro DLC campaigns do: **stable, hull-matched, memorable numbers** the crews can learn once.

## How

- **`game/data/carrier_comms.py`** — a curated per-hull plan table, keyed by pydcs ship type id:

  | Hull | TACAN | ICLS | Link 4 | ATC |
  |---|---|---|---|---|
  | CV-59 Forrestal | 64X `FID` | 9 | 336.9 | 304.0 |
  | CVN-71 Theodore Roosevelt | 71X `TRO` | 11 | 336.1 | 305.0 |
  | CVN-72 Abraham Lincoln | 72X `ABE` | 12 | 336.2 | 306.0 |
  | CVN-73 George Washington | 73X `GWN` | 13 | 336.3 | 307.0 |
  | CVN-74 John C. Stennis | 74X `STN` | 14 | 336.4 | 308.0 |
  | CVN-75 Harry S. Truman | 75X `HST` | 15 | 336.5 | 309.0 |
  | LHA-1 Tarawa | 41X `TAR` | 1 | — | 310.0 |
  | Kuznetsov (both) | 35X/36X `KUZ` | — | — | 311.0/312.0 |

  TACAN = hull number wherever that channel is legal for a surface T/R beacon (channels 2–30 and 47–63 X are excluded per the existing `UNAVAILABLE` table — hence Forrestal 59→64X and Tarawa hull 1→41X). Mod carriers without an entry keep the legacy allocator path end to end.

- **Precedence — the table provides defaults, never mandates.** Resolved in `GenericCarrierGenerator._resolve_{atc,tacan,link4,icls}`: a value stored on the control point (user-set from the base dialog, or persisted from an earlier turn) always wins; a curated channel something else owns falls back gracefully.

- **`TacanRegistry.alloc_near`** — when a map's *real* beacon owns the hull channel (Bagram is 74X on Afghanistan, Kandahar 75X), the boat walks outward to the **nearest valid free channel** (Stennis off Afghanistan gets 73X) instead of falling to 1X. `alloc_for_band` now marks issued channels in `allocated_channels`, which also fixes a latent cross-usage double-issue (the X-band T/R and A/A valid pools overlap at 37–46 and 100–126, and neither iterator marked what it handed out).

- **`IclsAllocator`** replaces the bare `iter(range(1, 21))` — curated picks and the sequential fallback share one pool, so two boats can never collide on an ICLS channel (previously a stored channel could be re-issued to a second boat).

- **Stability across turns** — every resolved value is persisted back to the control point (`frequency`, `tacan`, `tcn_name`, `link4`, `icls_channel`). Previously the ATC and the TACAN ident re-rolled every single mission and Link 4/ICLS re-allocated; now the squadron learns Mother's card once per campaign.

- **Flagship naming** — the page's Callsign line prints the flagship's *unit name*, so the carrier unit is named by its hull name ("CVN-74 John C. Stennis") instead of the id-prefixed theater-unit name. The name is set before `_register_theater_unit` records it, so debrief kill-tracking keys the same string; if a same-named unit already exists (two boats of the same class) it keeps the unique prefixed name.

- **Pretense**: keeps its own `generate()` behavior untouched — only the allocator type changed (`alloc()` is the same sequential 1–20 walk as the old iterator).

## Testing

- `tests/test_carrier_comms.py` (24 tests): table invariants (unique + T/R-legal TACAN, unique ATC/ICLS/Link 4, 336-band Link 4, 3-letter idents), the `IclsAllocator`, curated → stored → fallback precedence for all four resolvers, the nearest-neighbor TACAN degrade, and the flagship-name collision guard.
- Verified headless end-to-end through the real `GameGenerator → begin_turn_0 → MissionGenerator` pipeline on an Afghanistan campaign with a Stennis: the generated miz carries unit name `CVN-74 John C. Stennis`, TACAN **73X STN** (74X correctly ceded to Bagram's real beacon), ICLS **14**, Link 4 **336.4**, group ATC **308.000** — and the same values persist on the control point for the next turn.
- Black / mypy / pytest all green locally.

No settings, no save-format change: old saves keep any already-persisted channels by design, and fresh campaigns (or newly-taken carriers) pick up the curated card on their next generated mission.


BradySox:
Note on precedence, in case it comes up in review: a carrier TACAN stored on the control point (a user's base-dialog choice) always wins — `_resolve_tacan` only applies the curated card when the channel is unset — so this never overrides a hand-picked channel, and existing saves keep whatever the user chose.

(For context: the downstream fork this was carved from also persists an *auto*-assigned TACAN and adds a `tacan_is_auto` flag, so it needs an extra gate to let the curated card supersede a stale auto-pick from a pre-feature turn. Upstream has neither — it never persisted an auto TACAN — so the plain 'stored wins' rule here is already correct and no such gate is needed.)

PR #884, opened 2026-07-20
Fixed-wing air assault by CTLD paradrop (player + AI troop transports)

BradySox:
## Fixed-wing air assault by CTLD paradrop (player + AI troop transports)

**The gap:** Air Assault is *plannable* for the C-130 family — `Flight.is_hercules` covers both the Anubis mod and the official C-130J-30 module — but only the Anubis mod can actually deliver troops in flight, through its own loadmaster cargo system. The **CTLD troop pool itself can never leave a fixed-wing aircraft airborne** ("Too high or too fast to drop troops into combat"), so an official-module C-130J — and any **AI-flown** fixed-wing transport, the Hercules included — arrives over the assault zone with no way to unload. `C-130J-30.yaml` has carried `# Air Assault: 0 #TODO: Add once we have proper support for paradrops` since the module was added.

This PR is that support. It generalizes the existing Hercules planning branch to any troop transport and adds the missing delivery mechanism at the CTLD config layer (`CTLD.lua` itself untouched).

### Planner (`airassault.py`)
- The Builder gate becomes **"helicopter OR troop transport"** (`unit_type.cabin_size > 0`); a cabin-less fixed-wing still raises `PlanningError`.
- Any fixed-wing transport **preloads** (no pickup zone — the existing carrier/off-map/Hercules branch) and the CTLD assault-area waypoint becomes a **real AI run-in at 1,000 ft AGL** (`only_for_player = False` — exactly the existing Hercules shape, now generic). The Hercules keeps its `initial`-point ingress special case unchanged.
- `C-130J-30.yaml` gains **`Air Assault: 40`** — below the assault helicopters' 50, so a helo squadron in range still wins the tasking (the C-130 takes the long-reach and no-helo cases), and far below the Anubis mod's 990.

### Runtime (`ctld-config.lua`)
- The generator marks each troop-transport type **`paradrop`** in `Logistics.transports` (fixed-wing + `cabin_size > 0`, decided Python-side).
- **Player path:** while airborne, the stock CTLD F10 **"Unload / Extract Troops" jumps the stick** (`unloadExtractTroops` wrapped at the config layer, the same seam as the existing `ctld.inAir` CH-47 override). Grounded unloads, extraction, and **every helicopter path fall through to stock CTLD unchanged**. A player jump above **3,000 ft AGL** is refused with a message; the AI is exempt (its run-in is planned at 1,000 ft, and a terrain-forced high crossing must still deliver).
- **The drop:** the cargo leaves the aircraft immediately (weight adapted, coalition message) and the same troop group a landed unload would produce **ground-spawns at the velocity-projected drop point after a real static-line descent delay** (AGL ÷ 6.5 m/s, capped at 90 s) — so a transport shot down *after* the drop still delivers, and one shot down *before* never does. The landing reuses CTLD's own bookkeeping verbatim: `spawnDroppedGroup` (troops inside the active wpZone march to the zone centre — the existing air-assault behavior), the JTAC stick, the dropped-troop ledgers, `processCallback`.
- **AI path:** a 5 s release loop drops **one stick per sortie** when an AI transport crosses within 1,200 m of its **own** air-assault target-zone centre. Players are never auto-dropped; helicopters are never in the plan. This gives **AI Hercules assaults a working delivery for the first time** (the loadmaster system is player-driven); the Anubis mod's native cargo drops are a separate pool and are untouched.
- **Preload retry:** `preload_troops` retries every 30 s (~2 h give-up) instead of a one-shot at t+5 s — a TOT-delayed **late-activated transport no longer arrives over the drop zone empty** (fixed-wing has no pickup zone to self-load from).

### Tests
`tests/ato/flightplans/test_airassault.py` — the capability gate, the fixed-wing layout, the **helicopter layout pinned unchanged**, and the **Hercules layout shape pinned through the generalized path** (preloaded + `initial`-point ingress + 1,000 ft run-in).

Validated locally: Black clean, mypy clean, pytest 442 passed on `dev` @ `acf02b75`.

Downstream note: this is carved from the 414th fork (bradyccox/414Ret), where the same change ships with a lupa-harness runtime test suite for the Lua layer; happy to adjust anything for upstream taste.


BradySox:
@Starfire13

Druss99:
I think this needs to be play tested with all combinations. A mod C-130 player, mod C-130 AI (if possible I'm unsure), a base game C-130 player, and a base game C-130 AI

BradySox:
There are three C-130s in play, and one of the four combinations can't exist. Breaking it down:

| Airframe | What it is | Covered by this PR | Flown |
|---|---|---|---|
| `Hercules` (Anubis mod) | mod, player + AI | Yes | No — see below |
| `C-130J-30` (ASC module) | official module, player + AI | Yes | Player yes, AI not yet |
| `C-130` (base game) | AI-only, non-flyable | No | n/a |

Two notes on the matrix itself:

* A base-game C-130 player slot doesn't exist — `C-130` is a non-flyable AI unit.
* `C-130J-30` units ship with every DCS install, so the AI case runs for anyone. Only the player slot needs the module.

**The Hercules is not dropped by this PR.** The gate goes from `is_hercules` to `cabin_size > 0`, and `Hercules.yaml` already declares `cabin_size: 24` and `Air Assault: 990` — so it stays eligible and still outranks the C-130J-30's 40 by a wide margin. Both former `is_hercules` branches became `not is_helo`, which the Hercules satisfies, so its flight plan is byte-identical: preloaded, no pickup zone, assault-area run-in at 1,000 ft.

What is new for the Hercules is runtime behavior — the wrapped `unloadExtractTroops` and the AI release loop now apply to it. I have not flown that and I can't: my fork dropped the Anubis mod, so I have no install to test it on. That is the one part of your matrix I cannot close myself. If you or anyone running the mod can do a player unload and an AI sortie, that would cover it.

**What I have flown.** Player C-130J-30: the airborne "Unload / Extract Troops" menu drops the stick, the descent delay runs, and the troops land and behave. AI C-130J-30 is not flown yet — that is the run-in over the target zone plus the one-shot release loop, and it also exercises the preload retry against a TOT-delayed late-activated transport. I'll report back on it.

The Lua runtime is pinned by a headless harness on my side (player drop timing and projection, the jump ceiling, ground and helicopter fall-through to stock CTLD, the AI one-shot release, the preload retry). That is not a substitute for flying it, but it means the failure modes above are at least regression-guarded.

**Stock transports are now wired, pushed just now.** Your question exposed something bigger than the base-game C-130: every airframe the feature covered was western and module-or-mod gated. A sweep of every fixed-wing type declaring a Transport task found they all had the same defect — a Transport task and no `cabin_size`, so they defaulted to a zero cabin and the gate rejected them.

Five now carry a cabin of 24 and an Air Assault priority:

| Tier | Airframe | Air Assault | Note |
|---|---|---|---|
| strategic | Il-76MD | 15 | 22 red factions field it; the VDV's own jump platform |
| strategic | C-17A | 15 | 8 factions |
| tactical | C-47 | 20 | the 1944 factions' only option — they have no helicopters |
| tactical | An-26B | 25 | Soviet tactical hauler |
| tactical | C-130 | 30 | AI-only |
| tactical | C-130J-30 | 40 | unchanged |
| mod | Hercules | 990 | unchanged |

The tiering is deliberate: a tactical transport flies the low, slow run-in an assault wants and a strategic airlifter does not, so every tactical type outranks every strategic one. Within a tier the more capable airframe wins, and the player-flyable C-130J-30 beats the AI-only types. Where a faction fields only a strategic airlifter the number is moot, since it is the only option.

The cabin is 24 everywhere because that is the largest CTLD loadable group — a smaller number silently caps the aircraft at the 12-troop group. All five carried far more in service (An-26 about 40, C-17 102, Il-76MD 126).

The headline is the red side: those factions had no fixed-wing airborne option whatsoever before this.

Two things I deliberately did not do, both pinned by tests so they are not "fixed" later:

* The An-30M, C-2A and Yak-40 stay excluded. They declare a Transport task but are an aerial survey aircraft, a carrier onboard delivery aircraft and a light airliner. None carried paratroops.
* The A400M and V-22 are fielded by no faction, and the C-5 is strategic-only in practice, so none of them are worth the behavior change.

One consequence to flag, since it reaches past Air Assault: `LogisticsGenerator` also runs for TRANSPORT flights, so a positive cabin flips these types to `troops=true` in the CTLD transport table. They become troop-capable CTLD transports on Transport missions too. I think that is right for this set, but it is a behavior change on existing airframes, so say if you would rather any of them stay cargo-only and I will drop that commit.

---

### Separate finding: the Air Assault priorities need a rerank. Not in this PR.

Fitting five airframes into the list meant reading all of it, rotary included, and it is not a ladder — it is accumulated values, and several are wrong today on `dev`:

| Airframe | Now | Problem |
|---|---|---|
| `Hercules` | 990 | ranks a C-130 above **every** assault helicopter |
| `sh2f` / `vwv_sh2f` | 100 | naval ASW Seasprite, tied with the CH-53E |
| `SH-60B` | 70 | also ASW, above the assault UH-60A at 60 |
| `Mi-8MT` | 40 | the Soviet assault helicopter, in more red factions than any other rotary type, below a light observation helo |
| `Mi-26` | 30 | 24-troop cabin, below a UH-1H |

The Hercules value is the one that bears on this PR: `C-130J-30.yaml` says in its own comment that a transport should sit "below the assault helos so a helo squadron in range still wins the tasking," and at 990 the mod's C-130 does the opposite. These are not cosmetic either — `priority_list_for_task` feeds both `airwing.best_squadrons_for` and `squadrondefgenerator.generate_for_task`, so they pick the airframe that flies the assault and the one a def-generated squadron is built around.

What I would propose, banded rather than hand-set per aircraft:

| Tier | Band | Members |
|---|---|---|
| heavy-lift rotary | 110–135 | CH-47Fbl1 130 · CH-53E 125 · CH-47D 120 · Mi-26 115 |
| medium assault rotary | 80–105 | ch46d_late 100 · ch46d 98 · Mi-8MT 95 · UH-60L 92 · UH-60A 88 · UH-1H 85 |
| light / naval rotary | 50–75 | uh2c 70 · uh2b 68 · uh2a 66 · sh2f 62 · HKP15B 60 · SH-60B 58 · OH-6A 55 |
| fixed-wing tactical | 28–48 | Hercules 45 · C-130J-30 42 · C-130 38 · An-26B 34 · C-47 30 |
| fixed-wing strategic | 18–26 | C-17A 22 · IL-76MD 20 |
| gunship with troop seats | 8–15 | Mi-24P 12 · Mi-24V 10 |

Rotary transports beat fixed-wing ones, because a helicopter puts troops on the objective and can extract them while a paradrop is the long-reach answer; range is enforced separately in `can_auto_assign_mission`, so a transport still wins when no helo can reach. Gunships sit below the transports — six jump seats on an airframe that is a gunship first, and every faction fielding Hinds also fields Mi-8s.

**I have deliberately kept this out of this PR** — it is a helicopter rebalance, this PR is fixed-wing paradrop support, and the Hercules change is visible to every Anubis mod user. It is sitting on its own branch with tests pinning the bands and a guard that every Air Assault-capable airframe belongs to one. Tell me which you would prefer: fold it in here, take it as a separate PR later, or drop it entirely and I will leave the numbers alone.

Starfire13:
Couple things.

This needs testing with AI flights for sure. Not just player flown.

The Anubis mod already has its own implementation that works. This doesn't change that, correct?

I think the rerank thing is Claude overthinking things a bit. Maybe one or two of the chopper ranks needs tweaking. But not to that sorta formalised extent.

Will this implementation work with the C-47?

BradySox:
> Couple things.
> 
> This needs testing with AI flights for sure. Not just player flown.
> 
> The Anubis mod already has its own implementation that works. This doesn't change that, correct?
> 
> I think the rerank thing is Claude overthinking things a bit. Maybe one or two of the chopper ranks needs tweaking. But not to that sorta formalised extent.
> 
> Will this implementation work with the C-47?

Flown tonight and working with the ASC module, this does not change the Anubis module this is a CTLD modification. More spectator tests tomorrow.

Agrees on the rerank, I yanked it. I will always defer to you guys, just let me know.

C-47 works with this

PR #885, opened 2026-07-20
Custom victory conditions: authored win/lose blocks + domination/attrition options

BradySox:
## What

Alternate, legible ends to the war. The stock win condition is total conquest — `check_win_loss` returns WIN only when the enemy owns zero control points — which forces every campaign, including a limited war ("liberate the coast", a maritime pressure campaign), into a full ground invasion. This adds the community-requested shallow layer over that default, in two tiers over one engine (`game/victory.py`):

**Authored tier** — a campaign YAML `victory:` block declaring `win_when` / `lose_when` condition lists:

```yaml
victory:
  description: Liberate the coast
  win_when:
    - label: Take the coast
      capture_cps: [Sukhumi-Babushara, Gudauta]
    - enemy_air_below: 0.25   # attrit their air force to a quarter of start
      min_turn: 4             # guard: not before turn 4
  lose_when:
    - lose_cps: [Kutaisi]
```

Available condition fields: `capture_cps` (all named CPs blue) · `lose_cps` (any named CP red) · `territory_above` / `territory_below` (fraction of non-neutral CPs) · `destroy_targets` (all named TGOs fully dead) · `destroy_categories` (no red TGO of the class left alive) · `enemy_air_below` / `enemy_ground_below` / `friendly_air_below` (strength vs. the campaign-start baseline) · `enemy_air_denied` (no red CP with an operational runway) · `min_turn` guard · `label`.

**Generic tier** — two opt-in Settings knobs usable on ANY campaign with zero authoring (Campaign Management → Victory Conditions, both default 0 = off): **domination** (hold ≥ N% of the bases) and **attrition** (enemy air below N% of start, capped at 90 — "below 95%" would be a losses counter, not a victory condition). They synthesize into the same condition objects and stack with an authored block.

## Design decisions

- **A victory entry is a requirement, not a trigger**: EVERY field set on one entry must hold (AND within the entry); the `win_when` / `lose_when` lists are OR (any fully-met entry ends the war). That is what makes `min_turn` a guard ("not before turn 4") instead of nonsense ("win at turn 4").
- **Alternate endings ADD to the stock endings, never replace them.** Capturing everything still wins, losing everything still loses. Loss precedence on a simultaneous win+loss — a collapse is never a cheap win.
- **No vacuous wins.** Ratio conditions measure against a campaign-start `VictoryBaseline` latched in `initialize_turn` (getattr-guarded, so old saves latch on first load and need no migration); an empty baseline never fires. `destroy_categories` also requires the baseline to have counted at least one object of the class, so a category the campaign never fielded can't produce an instant win. A typo'd CP/TGO name can never be met — visible, never a silent win.
- **`enemy_air_denied`** counts `runway_is_operational()` — cratered airfields and sunk carriers are denied, FOB helipads count as air power, and a red off-map spawn is always operational, so the condition is unreachable on those campaigns by construction.
- **Announce once**: when a condition ends the war, `game.message()` names it ("Victory condition met — Take the coast") beside the generic Victory!/Defeat! dialog, latched per condition text because the UI flow calls `check_win_loss` several times around a turn boundary.
- **Zero planner coupling.** Ground truth, turn-boundary only; the AI commander never reads these conditions.
- Authored blocks are re-derived from the campaign YAML by name at load and never pickled; any lookup/parse failure degrades to "no profile" with a log (an old save whose campaign was removed still plays with the stock endings), while a broken block fails loudly in `parse_victory` so campaign CI catches it.

## Provenance

Running in the 414th fork (as fork feature §75) with the same engine + 30 of these tests; the fork additionally renders the conditions as a live checklist on its map-ribbon UI — `describe_condition(live=True)` ("Capture Sukhumi, Gudauta (1/2 held)") is the building block for that display and ships here, so a UI surface is a small follow-up if wanted.

## Testing

- `pytest tests` — 466 passed, 2 skipped (28 new tests: parse validation, every condition field, AND/OR semantics, baseline latch + no-vacuous-win guards, knob synthesis + stacking, loss precedence, the announce latch, and the real `Game.check_win_loss` branch order driven duck-typed)
- `mypy game tests` — no issues (490 files)
- `black --check` — clean

### DIFF
--- a/game/game.py
+++ b/game/game.py
@@ -56,6 +56,7 @@ if TYPE_CHECKING:
     from .sim import GameUpdateEvents
     from .squadrons import AirWing
     from .threatzones import ThreatZones
+    from .victory import VictoryBaseline
 
 COMMISION_UNIT_VARIETY = 4
 COMMISION_LIMITS_SCALE = 1.5
@@ -121,6 +122,12 @@ class Game:
         self.date = date(start_date.year, start_date.month, start_date.day)
         self.game_stats = GameStats()
         self.notes = ""
+        # Custom victory conditions: the campaign-start strength snapshot the
+        # ratio conditions measure against (latched by initialize_turn), and
+        # the once-per-condition announce latch. Absent on old saves; every
+        # reader guards with getattr, so no save migration is needed.
+        self.victory_baseline: Optional[VictoryBaseline] = None
+        self.victory_announced: set[str] = set()
         self.ground_planners: dict[UUID, GroundPlanner] = {}
         self.informations: list[Information] = []
         self.message("Game Start", "-" * 40)
@@ -420,6 +427,18 @@ class Game:
         persistency.autosave(self)
 
     def check_win_loss(self) -> TurnState:
+        # Alternate endings (custom victory conditions): one evaluator ahead
+        # of the stock capture-everything defaults. Returns None when nothing
+        # is configured, so this path costs nothing and the territory checks
+        # below remain the universal fallback.
+        from .victory import victory_verdict
+
+        alternate = victory_verdict(self)
+        if alternate == "loss":
+            return TurnState.LOSS
+        if alternate == "win":
+            return TurnState.WIN
+
         if not self.theater.player_points(state_check=True):
             return TurnState.LOSS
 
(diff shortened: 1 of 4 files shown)


Druss99:
I have a local branch for this already so if you don't mind I'll be taking this one 😅

BradySox:
> I have a local branch for this already so if you don't mind I'll be taking this one 😅

No worries, all yours. Any easy way to see who's working on what? I assume "issues" is that place?

PR #887, opened 2026-07-20
Soviet SHORAD batteries get their Sborka "Dog Ear" acquisition radar

BradySox:
Soviet-style SHORAD batteries currently deploy with no acquisition radar: an SA-9/SA-13/SA-15/SA-19 group generated from the generic SHORAD layout is launchers + logistics only, leaving the battery eyeball-only. In reality these systems fight as part of a regimental air-defense net cued by the MCC-SR Sborka "Dog Ear" — which is already a vanilla DCS unit with a unit yaml in the tree, just never placed.

## What this does

- The generic **`SHORAD.yaml` layout** gains an optional, non-filled **Search Radar slot** typed to the Dog Ear (one radar, placed with the battery).
- **`ForceGroup`** force-adds the Sborka into a group **only when all three gates pass**:
  1. a layout of the group actually declares a Dog Ear slot (`_layout_offers_dog_ear`) — no layout slot, no radar;
  2. the group is tasked **SHORAD** and its units match a curated Soviet-family marker list (SA-8/9/13/15/19, Osa, Pantsir, Strela, Tor, Tunguska) — a US Avenger or Roland battery on the same shared generic layout is untouched;
  3. the group doesn't already carry one (idempotent).

  The force-add is needed because factions don't list the Sborka directly, so `possible_types_for_faction` alone would leave the slot forever empty.

## What this deliberately does NOT do

- **SAM sites are untouched.** The Sborka is a regimental SHORAD acquisition radar, not part of SA-2/3/5 or S-300 site TO&E (and a mid-1980s anachronism at Vietnam-era SA-2 sites), so no SAM-site layout grows a Dog Ear slot — pinned by test.
- **Non-Soviet SHORAD untouched** (marker gate) — pinned by test.
- No faction file changes, no new units, no mission-runtime changes. Vanilla DCS unit only.

## Tests

`tests/armedforces/test_shorad_dog_ear.py` (new dir): SA-9/SA-15/SA-19 SHORAD presets gain the radar; Roland does not; SA-2/S-75, SA-3/S-125, SA-5/S-200 and SA-10/S-300PS presets do not.

Full suite: 441 passed, 2 skipped. Black + mypy clean.


red-one1:
DE for SA-15? Don't they already have search radars?

Druss99:
The force group changes shouldn't be needed, just the layout files and group files would need to be changed I think. 

Also I'm not sure this is needed. All of these have their own integrated radar IIRC and I've never seen them fire at significantly reduced range

BradySox:
> The force group changes shouldn't be needed, just the layout files and group files would need to be changed I think. 
> 
> Also I'm not sure this is needed. All of these have their own integrated radar IIRC and I've never seen them fire at significantly reduced range

Easy enough to yank the group layouts.

This PR is to give shorad groups DE search radar, which won't increase the range the missiles can reach, but it allows they to sit dark until que'd, which was unseen behavior before with Skynet

Ramius007:
AFAIK, those SR have no effect onany of mentioned systems, also some of them have own radar. EVEN if Dog Ear would help in target detection, it has so little range, that usually EWR would provide same benefit.

PR #888, opened 2026-07-20
Feature - Weapon standoff ranges

red-one1:
* **[Mission Generator]** Stand-off/cruise-missile-armed flights now ingress from a realistic launch distance. 
Weapons can declare a `range` (nautical miles) in their `resources/weapons` YAML, and when a package carries such a weapon whose range exceeds the doctrine ingress distance, t**he ingress point is pushed out to that range** (e.g. Tu-16s with Kh-22s begin their run ~160nm from the target instead of being dragged in to the doctrine ingress point). 
Weapons without a `range` are unaffected. 
Initial ranges are provided for the major stand-off, cruise, and anti-ship missiles and can be extended per-weapon.


BradySox:
Is the Auto-planner is still going to give cruise missile bomber flight a full escort package? What if we built in some kind of logic that if it is a cruise missile attack then that bomber flight doesn't need the whole SEAD/ESCORT package. Don't leave them unprotected unless the ingress is still under friendly barcap?

red-one1:
> Is the Auto-planner is still going to give cruise missile bomber flight a full escort package? What if we built in some kind of logic that if it is a cruise missile attack then that bomber flight doesn't need the whole SEAD/ESCORT package. Don't leave them unprotected unless the ingress is still under friendly barcap?

Haven't made any changes to the escort logic. Can consider it for later, but I don't want to bloat this PR too much.

geofffranks (approved):

game/ato/flight.py
```diff
@@ -289,6 +290,17 @@ def any_member_has_weapon_of_type(self, weapon_type: WeaponType) -> bool:
             m.loadout.has_weapon_of_type(weapon_type) for m in self.iter_members()
         )
 
+    def max_standoff_range(self) -> Optional[Distance]:
```
Druss99, line 294: If this is defined the same across all of the places we are using it I think we should factor it out into some sort of utility and import it everywhere instead of rewriting it

Druss99 (changes requested):
I think we need to test this pretty thoroughly as it could generate some very wonky flight paths or waypoints that are outside of the map area

game/ato/flight.py
red-one1: created game.utils.max_optional_distance

red-one1:
> I think we need to test this pretty thoroughly as it could generate some very wonky flight paths or waypoints that are outside of the map area

PackageWaypoints.doctrine_for_standoff_range now clamps the stand-off-driven max ingress distance to the departure-target distance

I will be testing this on my personal missions - I'll keep you updated

juanjux:
Testing this on my fork — three notes, no blockers (I still need to test this more, only tested it on one turn & mission).

- A clsid that isn't listed in its group's YAML silently gets no range: I hit
this with the LRASM: `AGM-158C.yaml` lists only `{B21_AGM-158C*4}`, so the single
`{B21_AGM-158C}` (declared by the CurrentHill USA pack, carried by the F-15EX mod which I have ported internally on my fork and I am testing)
lands in the "Unknown" group and the ingress stays at the doctrine distance. Same
for the 6x pack `{0290F5DE-014A-4BB1-9843-D717749B1DED}`, missing from `Kh-65.yaml`.
Nothing wrong with the logic, but it reads as "the feature doesn't work". Happy to
send those two clsids as a separate PR.

- Anti-radiation missiles are uncovered: Of the ARM-typed groups only the
bomber-launched ones (KSR-2P, KSR-5P, Kh-22P) got a range; HARM, AGM-88G AARGM-ER,
ALARM, Shrike, Standard ARM, MAR-1, LD-10, Kh-25MPU, Kh-31P and Kh-58U got none.
Deliberate scope, or a follow-up? AARGM-ER in particular exists to shoot from well
outside the doctrine ingress. Note that SEAD already shifts its hold point on
`any_member_has_weapon_of_type(ARM)`, so if ARMs ever get a launch range the two
would be moving the same thing.

- Nit: the ingress uses the package-wide maximum, so a package mixing a
stand-off shooter with a short-range flight pulls everyone out to the long range.
Intended, I assume, but it isn't stated anywhere.

One thing I checked so you don't have to: raising `max_ingress_distance` also feeds
`threat_tolerance`'s `target_size`, but it does not degrade IP threat avoidance —
with a 60nm SAM ring over the target the IP moves 45nm → 200nm and stays outside
the ring either way.

juanjux:
PS: tested on several missions, outside of the nitpicks above, it works well for Harpoons, Mavericks, LRASM, AGM-154* and other weapons I tried.

PR #890, opened 2026-07-20
Don't crash New Game on an authored-but-empty squadron aircraft: key

BradySox:
Defensive two-character guard for an easy campaign-authoring accident.

A squadron entry with an **authored-but-empty** `aircraft:` key —

```yaml
- primary: Transport
  secondary: any
  aircraft:
  size: 2
```

— parses the key as `None`, `data.get("aircraft", [])` passes the `None` straight through (the key exists, so the default never applies), and `DefaultSquadronAssigner` then iterates `config.aircraft` (defaultsquadronassigner.py:61) → `TypeError: 'NoneType' object is not iterable` **at New Game**, killing the whole campaign.

This is exactly what a campaign ends up with whenever an author removes a squadron's aircraft entries but leaves the key — e.g. stripping a mod aircraft out of a list (which is how we hit it on our fork: purging a mod plane from a campaign copy left `aircraft:` empty and New Game crashed).

## Change

`data.get("aircraft", [])` → `data.get("aircraft") or []`: an empty key now means the same as an omitted key — "any aircraft compatible with the primary task" (the existing `find_squadron_for` fallback).

Test: `tests/campaignloader/test_squadron_config.py` pins None → `[]` and that real preferences still pass through.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -25,6 +25,7 @@
 * **[Campaigns]** Ability to define motor pool objects which spawn reserve armor
 * 
 ## Fixes
+* **[Engine]** A campaign squadron entry with an empty `aircraft:` key no longer crashes New Game generation — it is treated as an empty list. (#890)
 * **[Mission Generator]** EWR sites now get the DCS "EWR" enroute task and come up on RED alarm, so their radars actually scan and report contacts (previously they could sit inert, especially with the "red alert state" performance option off). Works with or without the Skynet IADS plugin.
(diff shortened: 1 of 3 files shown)


Druss99:
If someone puts an aircraft key with no aircraft I think we do want to throw the error since they likely forgot to specify the aircraft

BradySox:
> If someone puts an aircraft key with no aircraft I think we do want to throw the error since they likely forgot to specify the aircraft 

Makes sense, close it or is there any usefulness we can take out of it?

PR #891, opened 2026-07-20
Load blue-block miz markers — 465 authored markers in 9 shipped campaigns never generate

BradySox:
`MizCampaignLoader` reads ships / offshore-strike / missile / coastal / LORAD / MERAD / SHORAD / EWR markers from the **RED country block only** — a marker authored in the BLUE block is silently dropped. A sweep of the shipped campaigns finds **465 authored blue-block markers across 9 campaigns that never generate**:

| campaign | dropped markers |
|---|---|
| normandy_full | **352** (SHORAD/missile/coastal belts) |
| normandy_small | **91** (75 SHORAD + 9 missile + 7 coastal) |
| operation_dynamo | 9 (the evacuation flotilla ships) |
| operation_allied_sword | 7 (oil platforms) |
| syria_full_map | 2 |
| golan_heights_lite / mozdok_to_maykop / RetakeTheFalklands / tripoint_hostility | 1 each |

(Marker units in the blue block — Tunguska/Scud/Silkworm placeholders on a WW2 map — are unambiguous authoring intent; the generator substitutes faction-appropriate hardware as usual.)

## What this does

1. **Every marker class also walks the blue block** (`itertools.chain(self.blue…, self.red…)`), the same way `aaa`/`armor_groups`/`helipads` already do.
2. **A blue-block marker prefers the nearest BLUE control point** — authoring in the blue block is an explicit ownership declaration, and nearest-any binding can hand the marker to the enemy field next door, where the objective then *silently never generates* (the owning faction has no ForceGroup for the other side's hardware; the only trace is a log line). Two guards keep this surgical:
   - **Scoped**: only the marker-class callers pass `prefer_blue` — the economy objects (armor/factories/ammo/strike) are *also* authored blue-block by convention and must keep binding by proximity, or red's economy would be re-owned to distant blue fields. (We shipped the unbounded version first on our fork and it re-owned 782 economy objects across the campaign set — Sperenberg's factory bound Frankfurt 408 km away. The scoping below is the corrected, verified shape.)
   - **Bounded** by `BLUE_BLOCK_MAX_DETOUR` (50 km): a legit near-field marker (Dynamo's flotilla, ~30 km) binds blue; a blue-block object sitting on an enemy base far from any blue field binds by proximity instead of being yanked across the map.
3. **Red-block markers are untouched** — the red block stays the coalition-agnostic default marker block binding by proximity, which is how campaigns conventionally author *blue* defenses near blue fields; binding red markers to red CPs would strip blue's defenses across the whole campaign set and is deliberately not done.
4. `generate_ewrs`' no-ForceGroup error now names the stranded marker, the bound CP, and its coordinates, so a mis-bound marker is diagnosable from the log line alone.

## Verified

- New integration test `tests/test_miz_marker_binding.py` (synthetic Caucasus miz built with pydcs): blue-block EWR planted 3 km from a red field generates and binds the blue field; red-block SHORAD next to a blue field still binds it by proximity; a blue-block **factory** on a distant red field binds the red field (scoping); a blue-block EWR beyond the detour bound binds by proximity; one within the bound binds blue.
- `normandy_small` headless end-to-end with the patch: `populate_theater` completes; the 75 SHORAD markers bind 54 blue / 21 by proximity; missile + coastal sites land on blue CPs.
- Full suite 440 passed, Black + mypy clean.

## ⚠️ One judgment call for you

The Normandy campaigns dominate the count (443 of the 465). Those markers are authored content that has plausibly been dead since the campaign-loader rework — resurrecting them is a real balance/density change for two campaigns that have been played for years without them (and Normandy was deliberately decluttered once, bfa0e4ba). If you'd rather keep the mechanical fix but not the Normandy resurrection, I'm happy to add a commit pruning/re-blocking those markers in the two Normandy mizzes — the fix itself is campaign-agnostic either way.

### DIFF
--- a/game/campaignloader/mizcampaignloader.py
+++ b/game/campaignloader/mizcampaignloader.py
@@ -189,45 +189,54 @@ class MizCampaignLoader:
             if group.units[0].type == self.NEUTRAL_FOB_UNIT_TYPE:
                 yield group
 
+    # Marker-block convention: the RED country block is the coalition-agnostic
+    # default marker block (campaigns author BLUE SAM/EWR sites as red-block
+    # markers near blue fields and proximity decides the owner), so red-block
+    # markers bind to the nearest CP of either side. A group authored in the
+    # BLUE block is an explicit blue-ownership declaration: it was silently
+    # dropped for the classes below (22 authored markers across 7 shipped
+    # campaigns never generated), and now binds with blue preference (see
+    # objective_info).
+
     @property
     def ships(self) -> Iterator[ShipGroup]:
-        for group in self.red.ship_group:
+        for group in itertools.chain(self.blue.ship_group, self.red.ship_group):
             if group.units[0].type == self.SHIP_UNIT_TYPE:
                 yield group
 
     @property
     def offshore_strike_targets(self) -> Iterator[StaticGroup]:
-        for group in self.red.static_group:
+        for group in itertools.chain(self.blue.static_group, self.red.static_group):
             if group.units[0].type == self.OFFSHORE_STRIKE_TARGET_UNIT_TYPE:
                 yield group
 
     @property
     def missile_sites(self) -> Iterator[VehicleGroup]:
-        for group in self.red.vehicle_group:
+        for group in itertools.chain(self.blue.vehicle_group, self.red.vehicle_group):
             if group.units[0].type == self.MISSILE_SITE_UNIT_TYPE:
                 yield group
 
     @property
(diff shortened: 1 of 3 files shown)


Starfire13:
352 EWRs in Normandy. Good lord. You could string them together across the channel and walk back to England.

What would actually be good is some consistency in the use of CJTF Blue and Red. For some objects, you can only use one, yet for others, both are acceptable. For example, SAMs have to be defined with CJTF Red, but AAA and static armour groups allow both CJTF blue and red.

PR #892, opened 2026-07-20
Expand SAM site layout variety and tighten the EWR radar pool (refresh of #791)

BradySox:
## SAM site layouts

- New dedicated **legacy battery sites**, each with proper search/track radar + launcher + point-defence + AAA composition (not just a launcher ring):
  - **SA-2**: 4- and 6-launcher × Circle/Semicircle
  - **SA-3**: 4-launcher × Circle/Semicircle
  - **SA-5**: Legacy Circle/Semicircle
  - **SA-6**: Reinforced Circle/Semicircle
  - **SA-2/SA-3 mixed site** (shared early-warning + both launcher types)
- The SA-2/3/5/6 groups now point at these dedicated batteries instead of the generic launcher rings.
- The existing Circle/Semicircle launcher sites are offered as options across the **modern** SAM batteries: SA-10/11/12/17/20/23, BUK-M3, HQ-22, Hawk, NASAMS, S-350, David's Sling, Iron Dome, Sky Sabre, THAAD.
- The new presets (notably the SA-2/SA-3 mixed site) are exposed in the relevant faction rosters.

## EWR radar pool fix

- Drop the `SearchRadar`/`SearchTrackRadar` fallback from the **Early-Warning Radar** layout. SAM search/track radars (Snow Drift, Straight Flush, Tin Shield, Fire Can SON-9, Silkworm SR, …) are SAM-system components, not standalone EWR sites, and were polluting the EWR buy menu.
- Reclassify the **P-14 "Tall King"** as `EarlyWarningRadar` — it's a dedicated long-range EW radar that belongs in this pool.

## Validation (re-run on current dev, 2026-07-19)

- Every preset group loads (68) and all **131 bundled factions** resolve with zero bad preset references — every group references a valid layout, every layout a valid unit, every faction roster a valid preset.
- Full suite: 438 passed, 2 skipped. No Python changes (resources + unit yaml only).


geofffranks:
Ran an ai review on this and it found a few things you might want to consider:

| Severity | Area | Issue | Suggestion |
|---|---|---|---|
| Important | `resources/groups/{BUK_M3,SA-11,SA-17,SA-6}.yaml` | `hide_on_mfd: true` is a **silent no-op**. The `ForceGroup` loader (`game/armedforces/forcegroup.py`) reads only `name/tasks/units/statics/layouts`, and `hide_on_mfd` is consumed only off the runtime `TheaterGroundObject` (set via the QGroundObjectMenu checkbox / migrator default `False`) — never from yaml. These SAMs will not be hidden on the F10/MFD. | Extend the loader + `ForceGroup` dataclass to read and forward `hide_on_mfd` into the generated `TheaterGroundObject`, or move it to the layout layer with loader support; otherwise remove the lines to avoid misleading config. |
| Minor | `Early-Warning_Radar.yaml` (+ `P14_SR.yaml` reclass) | EWR `fallback_classes` (SearchRadar / SearchTrackRadar) removed. Verified **safe** — no crash (`usable_by_faction` guards it) — but any faction with no `EarlyWarningRadar`-class unit now generates **no** EWR where it previously placed a search radar. | Confirm every faction expected to have EWR coverage still lists an EWR-class unit (1L13 / 55G6 / P-14 / Dog Ear / FPS-117 / FuMG-401 / FuSe-65). |
| Minor | `resources/layouts/anti_air/SA-2-SA-3 Mixed Site.yaml` | Internal slot/group labels are copy-pasted from S-300 (`S-300 Site SR1/CP/TR/LN1/LN2/SHORAD…`) for an SA-2/SA-3 site — functionally harmless but confusing to read/debug. | Rename to SA-2/SA-3-prefixed labels. |

BradySox:
Thanks — all three:

1. **`hide_on_mfd` no-op** — fixed by wiring it up rather than deleting it: `ForceGroup` now reads `hide_on_mfd` from the group YAML (explicit opt-in, default `False`) and forwards it onto the generated `TheaterGroundObject`, so the four mobile SAM sites (SA-6/SA-11/SA-17/BUK-M3) are actually kept off the F10/MFD datalink. Added a loader test. Kept minimal (explicit key + forward, not a broad task-based auto-hide — that's #794's territory).

2. **EWR `fallback_classes`** — good catch, and it's bigger than minor. I ran a faction-coverage sweep: **62 of 131 factions rely on the fallback** — they have no `EarlyWarningRadar`-class unit (russia_1980, iran_1988, all Israel/Syria, USA/USN, France, Sweden, most Cold-War reds), so removing it silently gives them **no EWR at all** and blinds their IADS (no crash, so CI misses it). I've **restored `fallback_classes`** — the EWR layout is usable by 106/131 factions again (the other 25 are WW2/insurgent factions with no radar at all, unaffected). Properly *tightening* the pool needs dedicated EWR units on those factions first (a separate content change), so this keeps coverage for now — which does mean the PR now mainly delivers the SAM layout variety, so it may be worth retitling.

3. **SA-2/SA-3 mixed-site labels** — I'm going to leave these. The mixed site reuses the shared `S-300_Site.miz` template (`layout_file:`), and the loader binds each slot `name:` to a group **name inside that template** (`group_for_name`), so renaming would break the mapping unless the site gets its own cloned template — a lot of surface area (and parking-fidelity risk) for a cosmetic rename you flagged as harmless. Happy to split the template if you'd rather.

Ramius007:
Sa-11/17 and propably Buk m3 as well, default layout is bad, for battalion sized unit it should be 6 launchers, 3 reloaders and SR, C2. It's standard russian unit, if You want to use smaller units, like battery, then just 2 launchers and reloader, without C2. Current 4 launchers layout not exist IRL in any known to me army.

PR #893, opened 2026-07-20
SAM guidance-radar redundancy — two engagement radars per site (builds on #892)

BradySox:
**Builds on #892** (the SA-2/3/5/6 battery layouts land there; this PR's first commit is that branch — review the last commit only, or merge #892 first and this rebases to a one-commit diff).

## The problem

Every SAM site layout fields exactly **one** engagement radar, so a single ARM on it is a functional site kill: launchers alive but blind, the site permanently out of the fight while still costing the AI (and the player) a full site's worth of units. In practice SEAD reduces to "one HARM per site."

## The change

Every SAM layout now fields **two guidance radars**:

- **Track Radar slots**: generic 2/4/6-launcher sites + SA-2 ×4 / SA-3 ×2 / SA-5 ×2 / S-350 / NASAMS-3 battery layouts (`unit_count` 1 → 2)
- **S-300 Site TR**: the S-300 family site + the HQ-22 battery
- **SA-2/SA-3 mixed site**: both channels (the Fan Song rides the CP slot)
- **SA-6 Reinforced**: the combined 1S91 STR (its Search Radar slot)
- **NASAMS 3 / Sky Sabre**: the Sentinel / Giraffe engagement radar in their dedicated layouts' Search Radar slot (AMRAAM/CAMM engagement stops without them)
- **Patriot / MIM-104 / LvS-103**: already fielded 2 STRs — now CI-locked

The shared `.miz` templates gained a second radar **position** per slot (`generate_units` hard-caps at the template's position count): 8/6-launcher circle + semicircle, 2_Launcher, S-300 site. New positions sit **45–121 m from the primary** (one ARM blast can't take both) and clear of every other unit; extra positions are inert for layouts keeping a lower `unit_count` (the S-300's own 54K6 CP stays 1).

Deliberate limitations:
- Presets routing a lone STR through a **generic** layout's Search Radar slot (NASAMS-B/C, IRIS-T SLM, THAAD) keep a single engagement radar — doubling that shared slot would double every generic site's pure search radars too.
- TELAR systems (SA-11/17, Roland, SHORAD) never had the single point of failure and are untouched.

## Why (and the honest caveat)

This is a **gameplay balance call, not a TO&E correction** — a real legacy fire unit fields one engagement radar. We wrote up the realism alternatives before choosing this:

- **Regiment-of-single-radar fire units** (kill one battalion's radar, the regiment fights on) is the historically faithful model for the strategic systems — and it already *emerges* in this engine by placing several single-radar sites on one objective and letting the IADS net them. Where a campaign authors that pattern, doubling should not stack on top (we run our S-300 regiments with the doubling reverted for exactly that reason).
- Revetment-geometry authenticity and radar decoys were considered and dropped (shared templates can't match every tenant's signature; ARM seekers home on emitters, so a functional decoy isn't buildable in DCS).

For the common case — one preset site defending an objective — two radars is the smallest change that makes SEAD a *campaign* against a site instead of one missile. Site price rises by one radar through the normal buy-menu math; buy maxes/defaults follow the layout automatically.

Happy to split any part out or park this one if you'd rather keep stock sites single-radar — it's the most opinion-shaped of the batch we're upstreaming this week, which is why the rationale is spelled out.

## Verified

`tests/armedforces/test_sam_radar_redundancy.py` CI-locks all 29 layout/slot pairs **both ways**: `unit_count == [2]` AND ≥ 2 template positions (so a yaml/template drift can't silently produce `LayoutException` at generation). Full suite 467 passed; resources + tests only, no Python changes.


Starfire13:
How do the AI handle the double radar? Particularly for LORADS where they have to perform saturation attacks. Does a single SEAD flight target both effectively?

geofffranks:
Ran an ai code review on this and it found things, but this was the only one that seemed like it might be something useful:

| Severity | Location | Finding & suggestion |
| :--- | :--- | :--- |
| Minor | `tests/armedforces/test_sam_radar_redundancy.py` | Contract is an **allowlist** of 29 known layout/slot pairs, so it won't fail if a *new* SAM layout ships with a single guidance radar. The deliberate single-radar exclusions (NASAMS-B/C, IRIS-T SLM, THAAD) are documented, so this is defensible — but consider a companion check that flags any `LORAD`/`MERAD`/`HERAD` layout whose guidance slot isn't `unit_count == [2]`, allowlisting those three. |

BradySox:
Good call — added a companion guard (`test_no_engagement_radar_layout_escapes_the_redundancy_contract`). Rather than extend the hand-list, it discovers every dedicated engagement radar from the layout data (a `TrackRadar`-class group, or a group whose units are in `radar_db.TRACK_RADARS`) and asserts each fields two — so a *new* single-radar SAM layout fails CI without anyone remembering to update `GUIDANCE_RADAR_SLOTS`. TELARs (radar == launcher) and generic Search-Radar slots are out of scope by construction; deliberate singles (the regiment-by-authoring pattern) are exempted by the "Single Radar" naming convention.

PR #896, opened 2026-07-20
Surface the squadron country: campaign yaml pin + Air Wing dialog selector

BradySox:
## What

Answers the Discord discussion about #854/#627: how do you choose a squadron's nation? The country
lives only in preset yamls, so an airframe-name squadron config under a CJTF faction picks its preset
by `random.choice` across every nation's presets — a USAF-named F-16 squadron can roll an Israeli
preset and fly the campaign with the wrong DCS voice and pilot names. The only fix today is
hand-authoring a preset yaml.

This PR surfaces the nation in both authoring layers.

1. **Campaign yaml `country:`** — a squadron config can pin its nation (`country: USA`). The pick then
   accepts only same-nation presets, falls through to the squadron-def generator when no same-nation
   preset exists, and `override_squadron_defaults` stamps the pinned country either way: a preset
   squadron for that nation if one exists, otherwise a generated squadron set to that country. An
   unknown `country:` name raises, so New Game aborts and names the bad value instead of silently
   flying under the wrong nation. Configs without the field are byte-identical to today.
2. **Air Wing Configuration dialog "Country:" selector** — under Livery. Opens on the squadron's
   current nation and writes `squadron.country` live, following the livery-selector pattern. The list
   is the full DCS country list, plus the squadron's current country when pydcs does not list it (a
   mod nation). Pilot names follow automatically: the wizard shows the dialog before
   `populate_for_turn_0` recruits the roster, and `Squadron.faker` (#854) reads the country live, so a
   mid-campaign change affects newly recruited pilots.
3. **Preset dropdowns show each preset's nation** (`VF-103 (Sluggers) [USA]`). Under a CJTF the
   presets span many countries and the country decides the comms voice.
4. **Save Config / Load Config round-trips the country** — `_build_air_wing` exports `country:`, so
   reloading an air wing no longer rerolls each squadron's nation. This records a `country:` for every
   squadron in the saved config.

**Bugfix in passing:** after "Replace with preset" the livery selector kept writing to the discarded
squadron object — `bind_data` re-bound the combo but never re-pointed `livery_selector.squadron`.
`bind_data` now re-points it. The country selector's `set_squadron` avoids the same trap.

## Changed on review

- **Operator-country trim removed** (Druss99, 31 Jul — "a massive burden when adding a new aircraft
  module"). `game/dcs/operatorcountries.py`, the curated operator tables and the operator-derived CJTF
  default are gone. An unpinned squadron under a CJTF keeps the picked def's own country: stock
  behavior, untouched by this PR.
- **Selector lists every DCS country again**, with no operator filtering.
- **An unknown `country:` raises** instead of degrading silently.

What is left is the yaml pin, the dialog selector, the preset-nation labels and the Save/Load
round-trip. No default behavior changes.

## Not included

- No shipped campaign uses `country:` yet. Behavior changes only when an author pins a nation or a
  player uses the selector.
- No Qt test — there is no `qt_ui` test precedent upstream. The pick and override semantics are
  covered by `tests/campaignloader/test_squadron_country_pin.py` (12 tests). Our fork additionally
  carries an offscreen-Qt selector test.

## Validation

- `black --check .`, `mypy game tests` and `pytest tests` on the branch. 

---
_Generated by [Claude Code](https://claude.ai/code)_


Starfire13:
Nice. I expect this will help with all the AI-only Redfor aircraft that have no preset squadrons (A-50, MiG-23, su-17, etc) getting set to the topmost livery (typically something that starts with A like Algerian) on campaigns where the faction is CJTF Red. Now I can explicitly set those squadrons as Russian in the campaign yaml.

BradySox:
> Nice. I expect this will help with all the AI-only Redfor aircraft that have no preset squadrons (A-50, MiG-23, su-17, etc) getting set to the topmost livery (typically something that starts with A like Algerian) on campaigns where the faction is CJTF Red. Now I can explicitly set those squadrons as Russian in the campaign yaml.

image
Can I interest you in a 3rd Reich Hornet? lol I'm trimming the country list right now, currently it just matches the whole available lot

BradySox:
image
Fixed, will now only show what the countries flew 
image
image
image

geofffranks:
I ran a code review through AI on this, and it came back with some minor issues you might want to address?

| Severity | Area | Issue | Suggestion |
|---|---|---|---|
| Minor | `AirWingConfigurationDialog._build_air_wing` | Saving config now writes `country` for **every** squadron unconditionally. This fixes the reload-reroll bug, but broadens `country:` from an author opt-in to always-on — a Save/Config pins every squadron's nation, including ones an author may have intended to roll randomly under CJTF. | Confirm this is intended; add a changelog/docs note that saved configs now pin the nation for all squadrons. |
| Minor | `operatorcountries.operator_countries` | A typo'd curated country name raises `KeyError` **inside the dialog's `__init__`** (crashes the Air Wing dialog on open), unlike `resolve_config_country` which logs and degrades. The CI test `test_curated_tables_resolve_against_pydcs` guards it, so practical risk is low — but the only safety net is CI. | Resolve curated names once at import (fail loud at startup) **or** degrade gracefully to mirror `resolve_config_country`; at minimum, document that CI is the guard. |
| Minor | `squadron.py:242-244` | Comment claims "the country/locale is fixed for a squadron's lifetime" — no longer true now that the dialog mutates `Squadron.country`. The code is still correct (the `faker` property re-reads `self.country`). | Update the comment so a future maintainer doesn't add a keyed-once faker cache that would silently break live country changes. |
| Minor | `tests/` | No test covers two **stated** requirements: (a) the Save↔Load round-trip of `country` (`_build_air_wing` → `from_data`); (b) the "named preset + different-nation pinned `country` → country re-stamped" behavior. Core assigner/operator logic is well covered. | Add focused tests for both paths. |

Starfire13:
> including ones an author may have intended to roll randomly under CJTF

I think for a faction that uses CJTF, not having a country defined for a squadron in the campaign yaml should either leave the country as CJTF by default or alternatively set it to the country of origin as defined in the aircraft yaml, rather than rolling randomly.

BradySox:
Thanks — addressed all four, plus @Starfire13's CJTF default:

1. **Save Config pins country for every squadron** — intended, and now noted in the changelog. It's also less broad than it reads after this change: an unpinned squadron under a CJTF faction no longer rolls a *random* nation (see below), so Save Config just records that already-deterministic nation rather than freezing a random roll.
2. **`operator_countries` KeyError** — now degrades: a curated name that doesn't resolve is skipped with a logged error instead of raising, so the Air Wing dialog can't crash on open. `test_curated_tables_resolve_against_pydcs` stays the loud dev-time guard, and there's a new `test_curated_typo_degrades_instead_of_crashing`.
3. **Stale `squadron.py` comment** — updated: the country is now mutable via the dialog, and the per-batch faker re-read is what keeps a live change working (with a "don't memoize it keyed once" note).
4. **Missing tests** — added `test_saved_country_round_trips` (export → `from_data`), `test_pinned_country_restamps_a_named_preset_of_another_nation` + `test_named_preset_pick_ignores_nation`, plus the CJTF-default suite and the degrade test.

@Starfire13 — implemented your default: an unpinned squadron under a CJTF faction now defaults to the airframe's most likely operator via `operator_countries` (A-50 → Russia, USAF F-16 → USA), falling back to the faction country when there's no operator data. Single-nation factions are unchanged.

game/campaignloader/defaultsquadronassigner.py
```diff
@@ -4,18 +4,50 @@
...
+#: a single-nation faction is already restricted to its own country.
+CJTF_COUNTRY_IDS = frozenset(
+    {CombinedJointTaskForcesBlue.id, CombinedJointTaskForcesRed.id}
+)
+
+
+def resolve_config_country(config: SquadronConfig) -> Optional[Country]:
+    """The campaign-pinned DCS country for a squadron config, if any.
+
+    An unknown name must never abort New Game: log it and return None so the
```
Druss99: I think we do want to abort the new game and throw an exception so the campaign owner knows they have something wrong in their campaign yaml

game/campaignloader/defaultsquadronassigner.py
```diff
@@ -55,36 +87,80 @@ def assign(self) -> None:
                 )
                 self.air_wing.add_squadron(squadron)
 
+    def _default_country_for(self, aircraft: AircraftType) -> Optional[Country]:
+        """The nation an *unpinned* squadron should fly under.
+
+        Under a CJTF/combined faction the airframe-name preset pick is otherwise
+        a random.choice across every nation's presets, so an A-50 with no preset
```
Druss99: I disagree here. We should generate a random squadron, if the random feature is broken and always selects the first item in the list we should fix that but hardcoding massive lists isn't ideal

Druss99 (changes requested):
This PR introduces a massive burden when adding a new aircraft module. My comments should help steer this a bit but TLDR; KISS, keep it simple. Allow country specs in the campaign yaml but don't adjust default behavior

BradySox:
> This PR introduces a massive burden when adding a new aircraft module. My comments should help steer this a bit but TLDR; KISS, keep it simple. Allow country specs in the campaign yaml but don't adjust default behavior 

I think Claude tackled both while I'm away from the PC, any other changes are easy just let me know

Druss99:
> > This PR introduces a massive burden when adding a new aircraft module. My comments should help steer this a bit but TLDR; KISS, keep it simple. Allow country specs in the campaign yaml but don't adjust default behavior 
> 
> I think Claude tackled both while I'm away from the PC, any other changes are easy just let me know 

Claude didn't update the PR 😂

BradySox:
> > > This PR introduces a massive burden when adding a new aircraft module. My comments should help steer this a bit but TLDR; KISS, keep it simple. Allow country specs in the campaign yaml but don't adjust default behavior 
> > 
> > I think Claude tackled both while I'm away from the PC, any other changes are easy just let me know 
> 
> Claude didn't update the PR 😂

Useless bum. I'll update when I'm at pc

BradySox:
Thanks — capitulated to all of this. Pushed in the latest commit:

- **Dropped the operator-country default entirely.** `game/dcs/operatorcountries.py` and its curated tables are gone (no more per-module burden), and `_default_country_for` is removed — an unpinned squadron under a CJTF faction keeps the picked def's own country, the stock random behavior. If the random pick ever misbehaves that's a separate fix, as you said.
- **Unknown `country:` now raises** so New Game aborts loudly and points the author at the bad value, instead of degrading silently.
- **Dialog selector** offers the full DCS country list again (no operator trim).

What remains is just the campaign yaml `country:` pin + the dialog selector + preset-nation dropdowns + Save/Load round-trip. Unpinned configs are byte-identical to before this PR. Tests + changelog updated; black/mypy/pytest green.

Issue #901, opened 2026-07-22
AAA sites are getting inappropriate search radars for their weapon types


red-one1:
### Affected versions

Test build

### Description

Possible AAA site:

<img width="313" height="391" alt="Image" src="https://github.com/user-attachments/assets/9d463ae2-6da4-444c-b5ca-74a1b410e770" />

The Patriot STR doesn't benefit the 88 guns at all - in this case, retribution should not create the searchradar.

### Save game and other files

.

red-one1:
I suspect this is because of the AAA_site.yaml

name: AAA Site
description: A standard AAA template
generic: true
tasks:
  - AAA
groups:
  - AAA:
    - name: AAA Site Radar
      optional: true # Optional if the AAA Site has a Radar
      unit_count:
        - 1
      unit_classes:
        - **_SearchRadar_**
    - name: AAA Site
      unit_count:
        - 2
        - 6
      unit_classes:
        - AAA
    - name: AAA Site Logistics
      optional: true
      unit_count:
        - 1
        - 2
      unit_classes:
        - Logistics

Starfire13:
Well also want to get Claude to have a look through all the current factions and add in EWRs to the ones that ought to have one but don't.

PR #902, opened 2026-07-22
only give guns a radar they can actually use (Fire Can ↔ KS-19)

red-one1:
## Symptom
Generic AAA sites were being generated with a radar the guns can't use — e.g. a SAM search radar (Patriot STR) sitting next to 8.8 cm Flak / S-60 guns. It contributed nothing to the guns and looked wrong on the map.

## Root cause
The generic `AAA Site` layout's optional radar slot accepted `unit_classes: [SearchRadar]`. Both the SON-9 "Fire Can" (an AAA gun-laying radar) **and** every SAM search/track radar share `class: SearchRadar`, so `TgoLayoutUnitGroup.possible_types_for_faction` would fill that slot with whatever `SearchRadar` a faction owned. With `fill` defaulting to `True`, `ForceGroup.for_layout` / `initialize_for_faction` auto-added that radar to generic AAA sites regardless of which guns filled the gun slot.

## Fix (minimal, data-driven)
- Added a dedicated `UnitClass.AAA_RADAR` and registered it in `ANTI_AIR_UNIT_CLASSES` (so `is_anti_air` stays true).
- Reclassified the two SON-9 Fire Can variants (`SON_9.yaml`, `Fire Can radar.yaml`) from `SearchRadar` → `AAARadar`.
- Changed the `AAA Site` layout radar slot to `unit_classes: [AAARadar]` **and `fill: false`**.

Net behaviour, matching the rule *"Fire Can only with KS-19; otherwise no radar"*:
- **Generic AAA sites** never auto-fill a radar (`optional + fill: false` is skipped by `for_layout` / `initialize_for_faction`) → guns spawn without a radar instead of with a useless one.
- **KS-19 / KS-19_HDS presets** explicitly bundle the Fire Can with their KS-19 guns, so at generation `dcs_unit_types_for_group` intersects the `AAARadar` slot with the preset's own units and the Fire Can appears **only** there.
- The `AAARadar` class also guarantees a SAM `SearchRadar` can never match the AAA radar slot even if bundled.

EWR sites are unaffected (they use the `EarlyWarningRadar` class/slot). WWII German flak still gets no radar — the Würzburg (`FuSe-65`) is classed `EarlyWarningRadar`, not an AAA gun-laying radar.

## Tests
Added `tests/data/test_aaa_radar.py` (revert-failing):
- Fire Can variants are `AAARadar` and still in `ANTI_AIR_UNIT_CLASSES`; Patriot STR stays `SearchRadar` (negative case).
- `ForceGroup.for_layout` on the generic AAA layout with a faction that has a Fire Can + non-KS-19 gun yields **no** Fire Can.
- A bundled KS-19 + SON-9 group still pairs them; a non-KS-19 bundle offers nothing to the radar slot.
- Guards the layout file: radar slot is `optional`, `fill: false`, `unit_classes: [AAARadar]`.

`pytest`, `black --check`, and `mypy` all pass.

## Compatibility
No save-format change. Only an enum value was added (`SearchRadar` retained), so existing saves unpickle fine, and `GroundUnitType.__setstate__` re-derives unit metadata from resources on load. Already-generated AAA sites in in-progress saves are untouched; every newly generated site follows the new rule. The layout pickle cache auto-invalidates via its source signature.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -25,7 +25,7 @@
 * **[Campaigns]** Ability to define motor pool objects which spawn reserve armor
 * 
 ## Fixes
-* **[Mission Generator]** Generic AAA sites no longer pull in a SAM search/track radar (e.g. a Patriot STR sitting next to 8.8 cm Flak) that the guns cannot use. AAA gun-laying radars (SON-9 "Fire Can") are now a dedicated unit class, and the AAA Site layout's optional radar slot only accepts them, so a faction without a suitable AAA radar simply gets an unradar'd gun line instead of a useless SAM radar.
+* **[Mission Generator]** Generic AAA sites no longer pull in a radar the guns cannot use (e.g. a Patriot STR sitting next to 8.8 cm Flak). The SON-9 "Fire Can" gun-laying radar is now a dedicated unit class and only appears where it is explicitly bundled with its guns (the KS-19 + SON-9 preset); generic AAA sites no longer auto-fill a radar, so a gun line without a matching radar simply spawns without one.
 * **[Mission Generator]** EWR sites now get the DCS "EWR" enroute task and come up on RED alarm, so their radars actually scan and report contacts (previously they could sit inert, especially with the "red alert state" performance option off). Works with or without the Skynet IADS plugin.
 * **[Plugins]** Fix the escort leash never running (DCS has no `Group.getByID`; look the group up by name via mist), so escorts are actually held to their engagement range.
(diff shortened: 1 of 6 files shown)


red-one1:
#901

geofffranks:
Some AI review feedback for you to consider: 

 > Consider adding a guard test that loads the real `resources/groups/KS-19.yaml` and `resources/groups/KS-19_HDS.yaml` and asserts each still lists a Fire Can variant. The new `AAARadar` class and its two reclassified unit yamls are reachable only through these presets — the current tests use synthetic `ForceGroup`s, so if the Fire Can is ever dropped from a KS-19 preset, `AAARadar` would become orphaned with no failing test to catch it.

PR #921, opened 2026-08-01
[FIX] Right-click on a front line under a flight-plan route opens the browser context menu

juanjux:
Right-clicking a front line that runs under a blue flight-plan route opened the browser context menu instead of the new-package dialog. Flight-plan routes draw a wide invisible hover polyline in the default `overlayPane`, on top of the front lines, so it took the `contextmenu` event.

Front lines now render into their own pane at z 450 — above `overlayPane` (400), below the marker pane (600). Panes hit-test in z order, so the front line gets the click back.

Test updated: the suite mocked `react-leaflet` without `Pane`, and it now also asserts the pane name and z-index.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,6 @@
 # Retribution v1.6.0
 
 ## Features/Improvements
-* **[Map]** Right-clicking a front line under a blue flight-plan route now opens the new-package dialog instead of the browser context menu (the route's invisible hover overlay swallowed the click).
 * **[Kneeboard]** Use a light-grey daytime kneeboard background instead of near-white, to avoid glare under HDR / Auto-HDR while staying readable in daylight.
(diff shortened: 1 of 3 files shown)


Starfire13:
Another long term issue fixed! Thanks!

This was a problem since Liberation and there wasn't a fix that worked so we thickened the frontline as a workaround.

Druss99 (approved):

changelog.md
```diff
@@ -1,6 +1,7 @@
 # Retribution v1.6.0
 
 ## Features/Improvements
+* **[Map]** Right-clicking a front line under a blue flight-plan route now opens the new-package dialog instead of the browser context menu (the route's invisible hover overlay swallowed the click).
```
Druss99: Can we put this under bug fixes

Druss99 (changes requested):

changelog.md
juanjux: Done!

Druss99 (approved):

PR #927, opened 2026-08-03
Cloud presets: selectable pack (Bandit / Weather 2.0 / ATMOS-X 1.2.0) + Live weather with ATMOS-X

juanjux:
Replaces the **Use Bandit's clouds** boolean with a **Custom cloud preset pack** dropdown (None / Bandit's Cloud Presets / Weather 2.0 / ATMOS-X), so the mission generator can use whichever community cloud-preset weather mod is installed.

### Live weather (ATMOS-X)

With the ATMOS-X pack selected, three extra settings appear (they stay hidden otherwise, since they can do nothing without it):

- **Use ATMOS-X live weather** — the turn's weather is a real METAR observation instead of a generated one, fetched with the ATMOS-X CLI.
- **CLI path** — blank by default; the CLI is found from the ATMOS-X installation registered with Windows.
- **METAR station (ICAO)** — blank by default; the station is picked from the airfields ATMOS-X knows on that terrain: the one the player is flying from if it reports, otherwise the nearest that does.

The observation is always the current one, and the mission always keeps its own date and time because the CLI serves 30 days of history at most.

If anything fails the turn keeps the weather Retribution generated.

### Weather panel

The turn's weather panel gained a tooltip carrying what its three words cannot: source (live METAR and the station, or a generated forecast), temperature, QNH in inHg/mmHg/hPa, turbulence, cloud preset with the layer breakdown from its description fog, visibility and the winds at all three levels.

With live weather on, a refresh button next to the panel re-fetches the METAR for the current turn, for the player who planned at dawn and takes off at noon, or whose first attempt found no network. Flight plans already built keep the altitudes they were given; the active runway and the carrier's course into wind are computed at generation and follow the new weather.


geofffranks:
Luna found one thing to address:

```
### Finding 1 — High: cloud-preset packs are not activated during new-campaign generation ###

File: `qt_ui/windows/settings/QSettingsWindow.py:404-420`

`_handle_mod_settings()` is only called from `QSettingsWindow.closeEvent()`. However, the new-campaign wizard uses `QSettingsWidget` directly (`QNewGameSettings.py:23-32`), and `NewGameWizard.accept()` immediately generates the game from those settings.

As a result, selecting Weather 2.0 or ATMOS-X in the new-game settings does not inject its `Preset35+` entries into `CLOUD_PRESETS` before weather generation. This means:

• generated campaigns do not actually use the selected custom pack;
• ATMOS-X live weather cannot resolve its reported cloud preset, because `_cloud_preset()` looks it up before the pack has been activated;
• the pack is only activated later if the user opens and closes the in-game settings dialog.

Activation should happen when the settings are applied/selected, or through a shared settings-application path used by both `QSettingsWindow` and the new-game wizard, before `GameGenerator.generate()` runs.
```

It seemed pretty certain, but this is how it would affect you if you want to double check.
 
```
The impact is primarily silent feature failure and incorrect weather, not usually a crash.

### Fresh application start ###

If the user selects ATMOS-X, Weather 2.0, or Bandit in the new-campaign wizard:

• the selected preset pack is not injected before generation;
• generated missions continue using whatever presets are already in `CLOUD_PRESETS`—normally stock DCS presets;
• the selected custom pack has no effect on generated cloud appearance.

### ATMOS-X live weather ###

For a live METAR that references a custom preset such as `Preset35`:

• `_cloud_preset()` cannot find that preset;
• the live weather model stores `preset=None`;
• the mission loses the intended ATMOS-X cloud texture/preset.

Clear observations using `"preset = 'none'"` are unaffected. Other METAR fields—wind, QNH, temperature, visibility, and precipitation—may still be applied, so this is not necessarily a total live-weather failure.

### If another pack was previously activated ###

The result can be worse: the process may still have an old pack injected. For example, selecting ATMOS-X in the wizard after Bandit was previously activated could cause ATMOS-X’s `Preset35` reference to resolve to Bandit’s cloud definition, producing the wrong cloud type rather than no preset.
```

Starfire13:
Is the Aurora automatically toggleable from outside the app, btw?

Right now, you have a choice of either cirrus clouds or Aurora as they use the same layer . And on missions where you want Aurora at night, you have to open the app and toggle Aurora on.

I was thinking it might be handy on missions that are set at night on maps that can experience autoras (Kola and South Atlantic). Not a necessary feature and barely a feature request tbh (it's not difficult toggling it in the app), more of a "I wonder if that's possible" thing.

juanjux:
@Starfire13 I will take a look at this, I think this could be doable with the command line which we call to retrieve the METAR preset anyway.

@geofffranks thanks, will check these as soon as I have some time.

juanjux:
> Luna found one thing to address:

Good catches, and it was worse, there was a third problem I only detected while debuggin this also affected loading a saved game. All should be fixed now, please tell this Luna thing to give it another pass.

Thanks!

juanjux:
> Is the Aurora automatically toggleable from outside the app, btw?
> 
> Right now, you have a choice of either cirrus clouds or Aurora as they use the same layer . And on missions where you want Aurora at night, you have to open the app and toggle Aurora on.
> 
> I was thinking it might be handy on missions that are set at night on maps that can experience autoras (Kola and South Atlantic). Not a necessary feature and barely a feature request tbh (it's not difficult toggling it in the app), more of a "I wonder if that's possible" thing.

Added two options below the atmos-x one to:
- Enable auroras at night on Kola or south pacific maps.
- Enable auroras at night on every map.

Starfire13:
Does this also then disable auroras on day turns? Otherwise Auroras get enabled on the first night turn and stay that way.

juanjux:
> Does this also then disable auroras on day turns? Otherwise Auroras get enabled on the first night turn and stay that way.

Yes, of course, only at night, with two notes:

- The Atmos-X setting for auroras is global, meaning, if you play a night turn with the autora setting on, dont finish the turn (force close retribution, it crashes, whatever) and then play a non retribution mission or campaign, auroras will be enabled until you disable them with the command line or the UI in Atmos-X. This is warned in the setting.

- They will be generated at night by the sun concept, not the current Retribution one which is not very good (that's a fix for another PR...), for example, yesterday I played a mission at 19pm on the 12th of december in the caucasus map, which DCS (correctly since dusk isat 17:43 that day) shows as close night, but Retribution still generated the light kneeboard.

juanjux:
@Starfire13 I am finding problems with the auroras support. I looks like the cli command only writes settings related to auroras or meteor showers but you need to actually open the UI for updating the shaders, so we can't really control it from our UI. Also, inconsistent results after even enabling or disabling it in the Atmos-X UI due to (I think) shaders cache. Removing the auroras support from the PR for now, the user still can enable or disable it manually with Atmos-X.

Starfire13:
That's fine. I was mainly curious if it could be done. At least it's not a hassle to toggle it manually. Thanks for trying though!

PR #929, opened 2026-08-07
Combat Search & Rescue (CSAR)

Drexyl:
Adds CSAR as a campaign-persistent mission type. An ejecting pilot becomes a persistent object on the theater map that survives across turns, can be targeted by a planned package, and goes MIA if nobody reaches them in time.

Features
Downed pilots persist across turns. An ejection leaves a survivor on the theater map with a survival clock (shorter near the front or in hostile territory); they go MIA if nobody reaches them. Shown as separate blue/red map overlays.

The autoplanner treats CSAR as a mission type. Plans rescue packages for both sides, AI or player-crewed, with escorts, and won't send helicopters into a live SAM ring. Capped per side per turn, and it won't waste a slot on a pilot it can't reach.

Any helicopter with a cabin qualifies, derived from the airframe — no configuration.

Rosters reflect it. A downed pilot holds their squadron slot; rescue them and they recover for a few turns, then return to duty.

Two pickup styles. Landing (the survivor walks aboard using DCS's own troop transport) or hover extraction (winched up by script). Survivors in the water always get the hoist, since nothing can land beside them.

Water rescues. A ditched pilot stays where they came down, anywhere on the map.

Survivors pop smoke for AI rescue flights; human crews get Ops.CSAR's beacons, MAYDAY calls and F10 menu and choose when to expose them.

Players can rescue anyone, in any CSAR-capable helicopter, including pilots assigned to an AI flight. Optionally requiring the cabin door open.

Rescues register at the right moment — an AI pickup when the survivor leaves the map, a player rescue when Ops.CSAR reports the disembark at base — and flow through state.json to the live Mission Status screen, the debrief and the roster.

Settings: CSAR on/off per coalition, ejection chance, survival and recovery turns, hover extraction, rescue AI pilots, require cabin door, dedicated CSAR start type, max CSAR flights per side.


juanjux:
Extremely cool. Extremely big to review too, I need a big cup of coffee :) However I will merge this into my personal fork and start to test in on my next campaign (I don't fly choppers for now so my testing will be limited).

PR #937, opened 2026-08-10
Update High Digit SAMs to 2.1.0

juanjux:
The New Game wizard still offered **High Digit SAMs (v1.4.0)** while the mod had moved on to 2.1.0. A two-way diff of the mod's `entry.lua` against our export showed drift in both directions: 11 ids we declare no longer exist, and 10 units the mod ships we never knew about. This PR closes both halves.

The SA-10B/S-300PS family has been added to DCS, so it's removed in this version. In some campaigns (e.g. Canary Islands invasion) this mean that if you had the updated mod, most of the group would appear on the UI but not on the mission. The preset now uses the stock S-300PS that ED added to base DCS.

Also retired: the orphaned `KS-19_HDS` preset. Nothing referenced it (the Egyptian and Iranian factions use the stock `KS-19/SON-9`, verified real) and both its units were dropped by the mod. The `Fire Can radar` and `KS19` classes stay declared but unreferenced so older campaigns still load.

On the other hand, some things where added:

The SAMP/T battery and the two early Strela MANPADS, with ranges read from the mod's own unit files:

| unit | range |
| --- | --- |
| SAMP/T MLT Launcher Block 2 | 200 km |
| SAMP/T MLT Launcher Block 1NT | 150 km |
| SAMP/T MLT Launcher Block 1 | 120 km |
| SAMP/T MRI Ground Fire 300 STR | 400 km detection |
| SAMP/T MRI ARABEL STR | 120 km, fire control |
| SAMP/T Control / Engagement / Generator | support |
| SA-7 Strela-2 / SA-7b Strela-2M | 3.2 / 4.2 km |

SAMP/T gets a preset and a battery layout (reusing `8_Launcher_Circle.miz`, like NASAMS-3) and goes to **Bluefor Modern**. It deliberately does *not* go to France 2005 or Italy 1990: SAMP/T entered service in 2010 and 2013, and shipping a unit a faction could not have is the same bug in a different coat. The Strelas go to Egypt 1973, Syria 1973, Iran 1988, Russia 1975/1980 and the GDR 1985 — six factions that until now fielded no MANPADS at all.


pydcs_extensions/highdigitsams/highdigitsams.py
```diff
@@ -3,6 +3,8 @@
 from game.modsupport import vehiclemod
 
 
+# Removed from High Digit SAMs in 2.x: DCS cannot resolve these two, so nothing may
```
Druss99: If we leave them here and leave the unit yamls then campaigns will spawn tanks instead of the SAM site. Instead we should migrate these units to the native DCS units like you did with the CH pack update

Druss99 (changes requested):

pydcs_extensions/highdigitsams/highdigitsams.py
juanjux: Yes, sorry, I started to work on this one before the CH pack update so yes, this need the same treatment, I will do it in two days (I'm outside home now). Good catch.

juanjux: Done. It was more than those two units. The whole S-300PS/SA-10B family was in the same state — CP, both TRs, both SRs, four launchers — 11 units in total, all of which ED had added to base
DCS: same treatment as the CH pack update: the 11 classes and their yamls are
gone, and both resolution paths now point at the native unit (the display-name
migrator in GroundUnitType, plus a new MigrationUnpickler handler for the pickled
classes) so a stale save resolves the same way either way. 

radar_db drops three
entries that were redundant anyway. 

One oddity left alone: our two TEL yamls carry the DCS TEL C/D names crossed
("S-300PS 5P85C ln.yaml" declares TEL D). The migrator targets whatever resolves
to the right id, with a comment. Happy to fix the swap separately in another PR if needed.

juanjux:
Closing this one, as I will make a new PR with these changes AND a selector to choose High Digit Sams Ultimate 1.4.4 on the new campaign dialog (and support for it, of course).

PR #938, opened 2026-08-11
Update CurrentHill Ukraine Asset Pack support

CleeeCATCH:
The pack renamed several of its DCS unit ids, so update pydcs_extensions to match and rename the resources that are keyed off those ids:

  Su-24MU        -> CH_Su-24MU
  MiG-29MU2      -> CH_MiG-29MU2
  BTR-4          -> CH_BTR-4
  T84_OplotM     -> CHAP_T84OplotM
  CH_T64BV       -> CHAP_T64BV

Aircraft icons and unit/payload yamls are keyed by dcs id, so those are renamed alongside. Also corrects the ukraine_2020 faction to reference the T-84 by its registered variant name, which otherwise raised a KeyError and stopped the whole faction from loading, and fixes the livery_name for the two renamed aircraft so their liveries resolve again.


game/factions/faction.py
```diff
@@ -940,12 +940,12 @@ def apply_mod_settings(self, mod_settings: Optional[ModSettings] = None) -> None
             self.remove_preset("Sky Sabre Battery (Mobile)")
             self.remove_preset("Stormer HVM Battery")
         if not mod_settings.ukrainemilitaryassetspack:
-            self.remove_aircraft("Su-24MU")
-            self.remove_aircraft("MiG-29MU2")
+            self.remove_aircraft("CH_Su-24MU")
+            self.remove_aircraft("CH_MiG-29MU2")
```
Druss99, line 944: Unit renaming has to be migrated, see past PRs on how to do this

Druss99 (changes requested):

PR #940, opened 2026-08-11
Add S-400/S-300V4

CleeeCATCH:
## Summary

Adds the **S-400 (SA-21 "Growler")** and **S-300V4 (SA-23)** to the HighDigitSAMs
integration. Both systems ship with the mod but weren't available in Retribution.

Adds two new preset groups — `SA-21/S-400` and `SA-23/S-300V4` — usable as LORAD
sites, alongside the existing SA-10/12/20/20B/23 family.

## What's included

**16 new ground units** in `pydcs_extensions/highdigitsams/highdigitsams.py`
with matching yamls in `resources/units/ground_units/`:

| S-400 (SA-21) | class | range |
| --- | --- | --- |
| 55K6 cp | CommandPost | – |
| 91N6E sr (Big Bird) | SearchRadar | 500 km |
| 96L6E sr / mast sr (Cheese Board) | SearchRadar | 330 / 520 km |
| 92N6E tr / mast tr (Grave Stone) | TrackRadar | 450 km |
| 51P6A ln (TEL A, 48N6DM) | Launcher | 250 km |
| 51P6A ln (TEL B, 9M96E2) | Launcher | 250 km |
| 51P6A ln (TEL C, 40N6E) | Launcher | 400 km |

| S-300V4 (SA-23) | class | range |
| --- | --- | --- |
| 9S457-2E cp | CommandPost | – |
| 9S15MDE sr (Bill Board) | SearchRadar | 500 km |
| 9S19M-1E sr (High Screen) | SearchRadar | 450 km |
| 9S32M-1E tr (Grill Screen) | TrackRadar | 400 km |
| 9A82M-2E / 9A84M-2E ln (Giant) | Launcher | 380 km |
| 9A83M-2E ln (Gladiator) | Launcher | 150 km |

**Other changes**

- `resources/groups/SA-21.yaml`, `resources/groups/SA-23_S-300V4.yaml` — new
  preset groups, both LORAD, both using the existing `S-300 Site` layout
- `resources/layouts/anti_air/S-300_Site.yaml` — new units added to the SR1, SR2,
  CP, TR, LN1 and LN2 slots. No new layout or `.miz` needed
- `resources/factions/{russia_2020,redfor_russia_2020,CH_russia_2020}.json` —
  both presets added
- `game/factions/faction.py` — both presets registered in the HighDigitSAMs
  mod-gate, so they're removed when the mod is disabled

## How the values were derived

Not estimated — taken from the mod's own data:

- **Radar `detection_range`** comes from each unit's `Range` in the HighDigitSAMs
  database. I calibrated this against five already-integrated units (PMU2 64H6E2 /
  92H6E, VM 9S15M2 / 9S19M2 / 9S32ME) and the mapping is exactly 1:1
- **Launcher `threat_range`** comes from the missile's `distanceMax`
  (`S_400_48N6DM` 250 km, `S_400_40N6` 400 km, `Antey4000_9M82MDE` 380 km,
  `Antey4000_9M83ME` 150 km)
- CPs and launchers use `detection_range = 0`, matching the existing entries

The S-400 TEL B carries a mixed 48N6DM/9M96E2 load, so it uses the longer 250 km
reach rather than the 9M96E2's 120 km.

Prices are scaled off the existing S-300 family (PMU2 CP 27 / LN 30 / SR 40 /
TR 33) with a small premium for reach — happy to retune if reviewers prefer
different values.

## Testing

- All 16 units resolve via `GroundUnitType.named()` with correct class and price
- Both groups load, resolve to the `S-300 Site` layout, and fill every
  non-optional slot; existing SA-10/10B/12/20/20B/23 groups still fill theirs
- Both presets load in all three factions and are correctly removed when
  HighDigitSAMs is disabled
- All 133 factions and every campaign yaml still parse


BradySox:
This is "High Digit Sam's ultimate" correct? The build of HDS I have does not have the SA-21, But the Ultimate collection does

CleeeCATCH:
> This is "High Digit Sam's ultimate" correct? The build of HDS I have does not have the SA-21, But the Ultimate collection does 

Roger

BradySox:
> > This is "High Digit Sam's ultimate" correct? The build of HDS I have does not have the SA-21, But the Ultimate collection does 
> 
> Roger

I'm m in favor of dropping HDS for HDSUC, and threw a PR up earlier for it but the two packs aren't compatible with each other and was causing issues- I think HDSUC is the better option, just gotta convince everyone else and or find a way to support both without errors

CleeeCATCH:
I agree, HDS Ultimate Compilation would be a better choice, it's more complete.

juanjux:
> > > This is "High Digit Sam's ultimate" correct? The build of HDS I have does not have the SA-21, But the Ultimate collection does
> > 
> > 
> > Roger
> 
> I'm m in favor of dropping HDS for HDSUC, and threw a PR up earlier for it but the two packs aren't compatible with each other and was causing issues- I think HDSUC is the better option, just gotta convince everyone else and or find a way to support both without errors

I was the one that tested both as being incompatible. I am also in favor of HDSUC is it's the better option, we just cant have both.

Druss99:
> > > > This is "High Digit Sam's ultimate" correct? The build of HDS I have does not have the SA-21, But the Ultimate collection does
> 
> > > 
> 
> > > 
> 
> > > Roger
> 
> > 
> 
> > I'm m in favor of dropping HDS for HDSUC, and threw a PR up earlier for it but the two packs aren't compatible with each other and was causing issues- I think HDSUC is the better option, just gotta convince everyone else and or find a way to support both without errors
> 
> 
> 
> I was the one that tested both as being incompatible. I am also in favor of HDSUC is it's the better option, we just cant have both.

Why are these incompatible exactly?

juanjux:
> > > > > This is "High Digit Sam's ultimate" correct? The build of HDS I have does not have the SA-21, But the Ultimate collection does
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > Roger
> > 
> > 
> > > 
> > 
> > 
> > > I'm m in favor of dropping HDS for HDSUC, and threw a PR up earlier for it but the two packs aren't compatible with each other and was causing issues- I think HDSUC is the better option, just gotta convince everyone else and or find a way to support both without errors
> > 
> > 
> > I was the one that tested both as being incompatible. I am also in favor of HDSUC is it's the better option, we just cant have both.
> 
> Why are these incompatible exactly?

@Druss99 don't remember the specifics of this one since I've been doing a lot of experiments lately, but I think the ultimate version reused some things from HDS for other sites which then did not work in mission.

Another option is to just let the user choose what they have, like I did in my PR for bandit/weather 2.0/atmos-x support. Note that HDS Ultimate is not from the same author so maybe some people could prefer the non ultimate version, which is probably more tested.

juanjux:
cc https://github.com/dcs-retribution/dcs-retribution/pull/956

Druss99:
Closing in favor of other PR

PR #949, opened 2026-08-19
Kneeboard: utc_kneeboard is read by nothing; print Zulu beneath local

BradySox:
`utc_kneeboard` is set on ten airframes and read by nothing. Their kneeboards print plain local times, with no Zulu anywhere, and have since the option was added.

Every airframe that declares it now gets each time's Zulu equivalent on a second line, in the flight plan and the support tables. The F-16C is added to that list.

```
Time
07:19:13
03:19:13Z
```

## The dead option

`generate_flight_kneeboard` converts the mission start time when the flag is set and passes the result to `BriefingPage` and `SupportPage` as `start_time`. Both store it. Neither reads it. `FlightPlanBuilder` takes a `start_time` too and never uses it.

Set on: `FA-18C_hornet`, `FA-18E`, `FA-18F`, `FA-18ET`, `FA-18FT`, `EA-18G`, and the six `VSN_F35` variants.

## Why one flag, not two

The earlier version of this PR added `annotate_zulu_time` and said it was independent of `utc_kneeboard`. Checking that claim is what found the dead option.

Both flags encode the same fact — this airframe's avionics run Zulu — so `utc_kneeboard` now drives the annotation and `annotate_zulu_time` is gone. The `start_time` parameters go with it; keeping one is what made the option look plumbed.

Nobody loses a local time they were coordinating on, because no Zulu time was ever displayed. The ten airframes gain the second line the option promised.

## Why the jet needs Zulu

System Time, shown on the DED and the HUD:

> Displays the internal system time in a 24-hour time format based on Zulu time (UTC).

— F-16C Early Access Guide p103, repeated p115.

The CRUS TOS DED page (p107) puts the desired Time-Over-Steerpoint beside System Time and derives the required ground speed from the difference.

## Why the second line rather than beside

The flight plan and support tables render through `tabulate` into a fixed-width image. An inline `07:19:13 (03:19:13Z)` on both the Time and Departure columns adds roughly 24 characters to a 9-column table, which risks clipping. Zulu underneath costs row height instead, which the page absorbs, and leaves every column width unchanged.

It also matches how the briefing page already shows two references:

```
Sunrise - Sunset: 05:42 - 20:11 (01:42 - 16:11 UTC)
```

## Testing

Seven tests. Five on the formatter (local, Zulu, the annotated pair, the double-print guard, `None`), and two at the table level where the defect was: a flight-plan row annotates when the airframe asks and stays bare when it does not.

459 passed, 2 skipped. Black and mypy clean.


Starfire13:
If there is going to be a change, it might be better to have the Viper's kneeboards have an additional time in Zulu? Because squadrons that fly multiple aircraft would be using the standard kneeboard time (i.e. not Zulu) to coordinate.

BradySox:
> If there is going to be a change, it might be better to have the Viper's kneeboards have an additional time in Zulu? Because squadrons that fly multiple aircraft would be using the standard kneeboard time (i.e. not Zulu) to coordinate.

Yep your right, pushed a update.

PR #952, opened 2026-08-21
Replace Bandit's cloud presets with ATMOS X

CleeeCATCH:
Replaces the Bandit's clouds mod support with [ATMOS X](https://www.digitalcombatsimulator.com/en/files/) v1.2.

ATMOS X occupies the same preset slots as Bandit's clouds (`Preset35` onwards), but ships 54 presets instead of 37 and renumbers several of them — `Preset40` is *Scattered Showers 6* under ATMOS X where it was *Overcast Light Rain 1* under Bandit's — so the preset table is regenerated from the mod's own `Config/Effects/clouds.lua` rather than relabelled. Names, METAR descriptions and altitude ranges come straight from that file, and the preset keys written to the `.miz` match it exactly.

### Changes

* **Setting**: "Use Bandit's clouds" becomes "Use ATMOS X clouds". Saves made before the swap carry the old toggle's value over (`Settings._migrate_legacy_cloud_mod`), and the unpickler maps the removed `pydcs_extensions.bandit_clouds` module onto the new enum so older saves keep loading.
* **Rainy presets**: mods can now declare which of their presets carry precipitation. pydcs' own rainy presets are identifiable by name (`RainyPresetN`), but modded ones have to keep the `PresetN` keys DCS expects, so `Clouds.random_preset` previously offered every modded rain preset — thunderstorms included — as "Cloudy" weather, and never used any of them for rain.
* **Registration timing**: presets were only injected when the settings window was closed, so a loaded campaign generated weather without them until the player happened to open settings. They are now registered from `Settings.apply_mod_cloud_presets()`, which also runs on campaign load and on new campaign creation.

### Testing

`tests/test_atmos_x_clouds.py` covers the preset keys, unique UI names, symmetric activate/deactivate, the rain/no-rain split and the settings migration. Verified against a live install as well: all 54 keys matched the installed `clouds.lua`, 6000 random draws stayed inside each preset's declared base range with the rain and dry pools disjoint, and a generated mission contained `["preset"]="Preset73"`. Full suite passes (456 passed, 2 skipped); black and mypy are clean on the changed modules.

### Note

This removes Bandit's clouds rather than offering both mods side by side, since the two occupy the same preset slots and cannot be enabled at once. Happy to restructure it as a choice between the two if you would rather keep Bandit's available.


Starfire13:
This is a duplication as Juanjux already has a PR awaiting review for this.

CleeeCATCH:
Got it

PR #957, opened 2026-08-29
[FIX] Building objectives that could never be recorded as destroyed

juanjux:
## The bug

Some building objectives would never be recorded as destroyed. An objective was only credited once every map object inside its trigger zone was dead, and many of those zones hold scenery that cannot be destroyed at all (`WOODPILE_01` and friends report a `getLife()` of `1e38`), so those objectives read as intact however often you flattened them, while others scored normally.

Nothing else caught them. DCS does report the death, but `getName()` on a scenery object returns the object's numeric id rather than a name, so the id went into `dead_events` and the debriefing, which resolves scenery by trigger-zone name, discarded it.

Over one Kola mission: 978 scenery deaths, 15 of them direct hits on named objectives. Two objectives were credited normally; three others recorded nothing at all across three turns while being levelled each time.

## The fix

Scenery deaths are matched to the nearest objective by position, in the base script.

The radius is measured rather than guessed: hits that destroyed the objective landed within **29 m** of its zone, collateral died from **31 m** out. 30 m keeps the first and rejects the second. Each objective is credited once, and unmatched scenery is dropped instead of appending a useless id, which also takes several hundred dead entries out of `state.json`.

Objectives already destroyed are seeded too, marked `dead`. They start pre-counted so they are never scored again, but they have to be in the list: the destruction zone that replays their rubble at mission start kills their scenery, and those deaths would otherwise be credited to whichever live objective happened to be nearest.

The `MapObjectIsDead` triggers go with it — 342 of them in one mission.

## Verified in game

The three objectives that had recorded nothing across three turns:

| objective | result |
| --- | --- |
| 4-building port | **4 / 4** destroyed |
| 6-building factory | **6 / 6** destroyed |
| 10-building camp | **9 / 10** destroyed |

Matching the log building by building and the save afterwards, against 667 pieces of collateral correctly rejected and no false positives. The one camp building the log did not credit is the one still standing in the save.

## Notes for review

- `SCENERY_MATCH_RADIUS` is the one tunable. The tightest successful match observed was 29 m, so it should not go below 30.
- A `_CRASH` (rubble) model dying also credits the objective. At 0–1 m it is the same spot and the dedupe stops it counting twice, but it is worth an explicit decision.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,7 @@
 # Retribution v1.6.0
 
 ## Features/Improvements
-* **[FIX]** Some building objectives would never be recorded as destroyed. An objective was only credited once every map object inside its trigger zone was dead, and many of those zones hold scenery that cannot be destroyed at all, so those objectives read as intact however often you flattened them, while others scored normally. 
+* **[FIX]** Some building objectives would never be recorded as destroyed. An objective was only credited once every map object inside its trigger zone was dead, and many of those zones hold scenery that cannot be destroyed at all, so those objectives read as intact however often you flattened them, while others scored normally. Objectives standing on scenery with no damage model, such as the moored submarines on Kola, are now credited on a direct hit.
 * **[Modding]** Added support for the CurrentHill Iran Military Assets pack: the Shahed-136 launcher, two IRGCN fast-attack craft, and a new `[CH] Iran 2020` faction, behind a New Game mods checkbox. (#886)
 * **[Kneeboard]** Use a light-grey daytime kneeboard background instead of near-white, to avoid glare under HDR / Auto-HDR while staying readable in daylight.
 * **[UX]** Hovering a friendly flight's route line on the map highlights it in yellow, and clicking it selects that flight's package (and the flight) in the ATO sidebar.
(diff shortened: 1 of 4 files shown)


Starfire13:
Just in case this is a concern, we originally used object IDs to determine deaths, but the object IDs would change any time the map got updated, which necessitated re-creating all the map object strike targets in any campaign on that map. So that got changed to the current system. From a quick read, I don't think your improvement breaks this, but just wanted to be sure I'm not missing anything and that you were aware that the old object ID-based system is not viable.

juanjux:
> Just in case this is a concern, we originally used object IDs to determine deaths, but the object IDs would change any time the map got updated, which necessitated re-creating all the map object strike targets in any campaign on that map. So that got changed to the current system. From a quick read, I don't think your improvement breaks this, but just wanted to be sure I'm not missing anything and that you were aware that the old object ID-based system is not viable.

It isn't since the id is only used to tell scenery from a unit, then discarded; what goes into dead_events is the trigger zone's name, same as today, and positions come from the campaign's own zones rather than from DCS.

If anything it's the opposite: before this, the numeric id was being appended to dead_events (~400 entries in one mission), so this removes the last remnant of the id-based path.

PR #961, opened 2026-08-31
Update Autoplanner for Motorpool Attacks

geofffranks:
Updates the autoplanner to assign armed recon attacks to motorpools, and bumps it up in the order of preferred targets when autoplanning.


Starfire13:
Do we have a "minimum number of motor pool targets to bother planning a mission", like we do for OCA/Aircraft? 4 is probably a decent minimum.

PR #966, opened 2026-09-12
Add native DTC data cartridges for the F/A-18C and F-16C

BradySox:
image

Every blue client flight of an F/A-18C or F-16C spawns with a native Data Transfer Cartridge auto-loaded. The cartridge is a JSON file at `DTC/<name>.dtc` inside the miz plus a per-unit `DTC` block, so multiplayer clients receive it with the mission download.

**Depends on dcs-retribution/pydcs#39** (the miz and unit seams). `requirements.txt` pins that PR's head commit; the pin moves to the merge commit when it lands.

## What the cartridge carries

| | F/A-18C | F-16C |
|---|---|---|
| Route | steerpoints with per-leg altitude, speed and ETA; STPT n is kneeboard waypoint n | steerpoints 1-20 with TOS and IP/TGT sub-types (the jet auto-sequences 1-20 only) |
| Recovery | TACAN tuned to the departure field on a land start when it has one, else the recovery field's; the carrier's card on a boat; ICLS/ACLS from the recovery; FPAS home waypoint; the bullseye as the A/A waypoint | DEST 81-99: the briefed divert first, the enemy field the flight is working over (within 10 NM of the target) next, then the nearest fields |
| SA / HSD | front line(s) as FLOT lines; the flight's own orbit (or its hold point) as CAP point 1, then the tanker and AWACS orbits; enemy SAM sites as MEZ rings | front line(s) as HSD lines; own orbit and tanker/AWACS orbits as steerpoints 21-25; enemy SAM sites as threat points (Custom type, radius from the site's threat range) |

Times are seconds since Zulu midnight: the cartridge clock is Zulu, not the local mission clock. Radio presets are not written: they already reach both jets through the miz.

## Scope

- One builder per descriptor (`hornet.py`, `viper.py`), shared extraction in `common.py`, the pass in `generator.py`. Per flight, not per type: each flight flies its own route.
- Best-effort: a failure building one flight's cartridge skips that flight; a failure in the pass never blocks the mission save.
- The cartridge is an overlay. Radio presets and the route already reach the jet through the miz, so with the setting off nothing changes.
- `DtcOptions` carries per-flight section switches, all on; only the campaign setting `Pre-load DTC data cartridges` (default on) is surfaced. A per-flight UI can be added without touching the builders.
- A steerpoint's `alt` is the altitude the miz route gives the jet: the planned altitude on an en-route point (the DED shows this field as ELEV; flown), the field elevation at takeoff and landing. Known gap: a point the miz puts on the deck (targets, flyovers) or plans AGL reads 0, because the generator has no terrain height source; the route entry keeps its AGL encoding for the DTC Manager.
- Not included: the F-14B(U) (a different descriptor schema), the CJS Super Hornets (their descriptor has no SA table), and the F-16C CMDS and ROE partitions.

## Verification

- Formats confirmed against the mission editor's own descriptors (`CoreMods/aircraft/<type>/DTC`) and against cartridges authored in the editor.
- Flown in both cockpits on the 414th's fork of Retribution: the cartridge auto-loads at spawn, and the route with TOS/ETA, the recovery aids, the A/A waypoint and the SA/HSD picture read as written.
- `black --check .`, `mypy game`, `mypy tests` and `pytest tests` (486 passed, 2 skipped) on this branch with pydcs#39 installed.


BradySox:
@juanjux any additions you wanna see? Ideas? Still needs the pysdcs merge first but this should be a easy project

juanjux:
Nice! I added a feature on my fork that also uses DTC, allows you to pick any point or unit in the map as a new waypoint (stored in a different SEQ or FPLAN) and has a new dialog with the added waypoints. I use DTC for adding the waypoints for the planes where it is supported and plain mission editor waypoints in others like the A-10. But it doesn't support so many things on the DTC itself.

BradySox:
@juanjux Thanks. Your added-waypoints feature sounds like it belongs next to this rather than inside it, so I've left it out here.

This PR only writes route sequence 1, so sequence 2 is free for your points on both jets:

- Hornet: `NAV_ROUTE[0]` only, `R2` false on every point.
- Viper: `R1` only. STPT 1-20 are the route, 21-24 the support anchors, and 25 is left alone because the jet loads the bullseye there.

If yours lands first I'll rebase this around it.

